PR #13, opened 2017-07-16
Add wagtail

marcoacierno:


backend/requirements/base.txt
```diff
@@ -1,3 +1,4 @@
-Django==1.11
```
marcoacierno: The version of wagtail I used when I started the integration wasn't compatible with django 1.11 yet

backend/requirements/base.txt
```diff
@@ -2,3 +2,4 @@ Django==1.11.3
 wagtail==1.11.1
```
marcoacierno: why?

backend/pycon/settings/base.py
```diff
@@ -136,7 +136,7 @@
 MEDIA_URL = '/media/'
```
marcoacierno: done

backend/requirements/base.txt
marcoacierno: maybe the comment is for 'psycopg2==2.7.1' and not wagtail?

patrick91: I'd just use postgres everywhere :)

simobasso: 'cause the actually default is `sqlite`.

If a collaborator w/o pg start an instances with that commit, the setup will fail.

options:

1. move `psycopg2==2.7.1` in production' dependencies
2. use pg as default in settings_base.py

patrick91: let's use postgres everywhere :)

PR #31, opened 2017-07-19
Remove required secret key for tests

patrick91:
I'm doing this to prevent issues when testing PR like in #28

But I think it will fail anyway since we are using secrets[1] for codacy. Maybe we can skip uploading coverage to codacity when testing external PRs. @simobasso ?

[1] Circle CI allows to not pass secrets to forked PRs build, since it is easy for someone to get access to them

### DIFF
--- a/backend/pycon/settings/test.py
+++ b/backend/pycon/settings/test.py
@@ -1,4 +1,4 @@
 from .base import * # noqa
 
 
-SECRET_KEY = env('SECRET_KEY')
+SECRET_KEY = 'this-key-should-only-be-used-for-tests'


simobasso:
IMHO tools like codacy are useful especially on pr (simplified review). What is the risk if anyone, stole that api-key? lie on our coverage? I think we can bear it :P

patrick91:
> IMHO tools like codacy are useful especially on pr (simplified review). What is the risk if anyone, stole that api-key? lie on our coverage? I think we can bear it :P

It is not about that specific API key, the issue is with other API keys, especially if we enable continuous deployment. Image if we use AWS, someone could easily get access to the AWS_ACCESS_KEY and secret :)

patrick91:
An alternative would be to have the codacy key in the config file, so that it is always available :)

PR #36, opened 2017-07-21
social login

patrick91:
I've used [Python Social Auth for django](https://github.com/python-social-auth/social-app-django).

The idea is that the frontend will open a popup to **/login/twitter** or any other social link, then the application will redirect to a post login view that will send a message (just a function call atm) that will let the frontend know that the user has been authenticated successfully or not.

# TODO: 

- [x] pass user information to frontend
- [x] add username to user, required by social auth


simobasso:
LGTM

patrick91:
@simobasso can you review this :)

PR #79, opened 2018-01-20
[WIP] Add initial stripe integration

patrick91:
Add initial implementation of stripe, stores customer id on the User model and will add some utility methods to fetch orders, cards, etc


patrick91:
Closing in favour of another approach

PR #177, opened 2018-05-26
Use material components web/react

patrick91:
As discussed with @fedriz and maybe others, we are trying to use the new [material design library](https://material.io/develop/), there are some issues with the [react library](https://github.com/material-components/material-components-web-react) so for now the idea is to vendorise the packages and use them directly with the material components web css.

Issues:

- We need to manually sync the packages when there's an update
- We are including the whole material design css
- We can use theming

All these issues will be fixed when the react library will be in a beta or production stage, hopefully this will be soon enough, but for now I guess having the components in our repo is not too bad


patrick91:
I think as a first step this is ready to be merged in, we can update the rest later (like the cards)

PR #351, opened 2018-09-22
WIP: Implement users query

marcoacierno:


backend/tests/users/test_schema.py
```diff
@@ -28,3 +31,172 @@ def test_works_when_user_is_logged_in(user, graphql_client):
...
+
+    assert "errors" in resp
+    assert resp["errors"][0]["message"] == "You don't have the permissions"
+    assert resp["data"] is None
+
+
+def test_query_all_users(admin_user, user_factory, graphql_client):
+    graphql_client.force_login(admin_user)
+
+    users = [user_factory() for i in range(0, 5)]
```
patrick91: this can be just range(5) :)

backend/tests/users/test_schema.py
```diff
@@ -28,3 +31,172 @@ def test_works_when_user_is_logged_in(user, graphql_client):
...
+    } == resp["data"]["users"]["objects"][1]
+
+
+def test_request_users_with_limit_outside_allowed_value(admin_user, user_factory, graphql_client):
+    graphql_client.force_login(admin_user)
+
+    users = [user_factory() for i in range(0, 100)]
+    users.append(admin_user)
+
+    resp = graphql_client.query(
```
patrick91: can you move this to a function? so you don't have to repeat the query so many times :)

frontend/src/components/pagination/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <div className={styles.pagination}>
+        <span
+          className={prevClasses}
+          onClick={this.changePage.bind(this, currentPage - 1)}
+        >
+          Prev
+        </span>
+        {Array(totalPages)
+          .fill(null)
+          .map((_, i) => i + 1)
```
patrick91: I'd put this in a utility method, mainly to make it more readable :)

`range(totalPages).map()`

PR #366, opened 2018-10-17
Conference

marcoacierno:
Base PR for the conference features

"Sub tasks" PRs:

- CFP: #374
- Voting
- Tickets


marcoacierno:
Let me know what do you think so far of the code @patrick91  :)

backend/languages/languages.py
```diff
@@ -0,0 +1,7 @@
...
+'''
+
+LANGUAGES = json.loads(_codes)
```
patrick91: I'd add just a few language for now, like Italian and English, no need to populate all the languages of the world :)

marcoacierno: done! :)

marcoacierno:
@patrick91 we have the `Ticket` model here, but do we need it considering we will using ti.to? To list the tickets in the frontend we can implement it fetching the tickets from ti.to (and caching them to avoid problems)

PR #374, opened 2018-10-27
CFP

marcoacierno:
Features:

- [x] Propose talk, workshop etc.
- [ ] Throttle requests?
- [ ] ?

Depends on #366


marcoacierno:
@patrick91 I was thinking to rename `Talk` to `Submission` because I think it makes more sense (you can submit a workshop, tutorial, talk, and other stuff to a Conference, and all of them share the same information and are pretty much the same thing). 

To identify the type we can add a field named `type` with the various types like:

- talk
- workshop
- tutorial

So something like this:

**Submission**

- conference `models.ForeignKey`
- title `models.CharField`
- abstract `models.TextField`
- elevator_pitch `models.TextField`
- notes `models.TextField`
- speaker `models.ForeignKey`
- topic `models.ForeignKey`
- language `models.ForeignKey`
- type `models.ForeignKey` (to = `SubmissionType`)

**SubmissionType**

- name `models.CharField`

Examples of Submission types:

- Talk
- Workshop

The mutation might be renamed to `sendSubmission` instead of `proposeTalk`.

`Conference` would have a `allowed_submissions` field to specify which kind of submissions are allowed.

marcoacierno:
Example query:

```
mutation {
  sendSubmission(input: {
    title: "Test title", 
    abstract: "This is an abstract", 
    topic: 1, 
    language: "it", 
    conference: "pp-18", 
    type: 2, 
    duration: 1, 
    elevatorPitch: "test pitch"
  }) {
    submission {
      id
      title
      abstract
      topic {
        name
      }
      type {
        name
      }
      duration {
        name
        id
      }
      elevatorPitch
      notes
    }
    errors {
      field
      messages
    }
  }
}

```

backend/api/forms.py
```diff
@@ -0,0 +1,7 @@
+from django import forms
+
+
+class GrapheneModelForm(forms.ModelForm):
```
patrick91: can you add a doc string here to make clear what this is for? :)

backend/conferences/models.py
```diff
@@ -15,6 +15,17 @@ class Conference(TimeFramedModel, TimeStampedModel):
...
     audience_levels = models.ManyToManyField('conferences.AudienceLevel', verbose_name=_('audience levels'))
+    submission_types = models.ManyToManyField('submissions.SubmissionType', verbose_name=_('submission types'))
+
+    @property
+    def is_cfp_open(self):
+        try:
+            cfp_deadline = self.deadlines.get(type=Deadline.TYPES.cfp)
+
+            now = timezone.now()
+            return cfp_deadline.start <= now <= cfp_deadline.end
```
patrick91: I like this, I was thinking that we should probably add the timezone to the conference, so that we can have conf in multiple timezones :)

backend/submissions/models.py
```diff
@@ -0,0 +1,91 @@
...
+        on_delete=models.PROTECT
+    )
+
+    duration = models.ForeignKey(
+        'conferences.Duration',
+        verbose_name=_('duration'),
+        on_delete=models.PROTECT
+    )
+
+    def clean(self):
```
patrick91: 👏

backend/tests/conferences/factories.py
```diff
@@ -66,6 +66,34 @@ def languages(self, create, extracted, **kwargs):
             for language_code in extracted:
                 self.languages.add(Language.objects.get(code=language_code))
 
+    @factory.post_generation
+    def submission_types(self, create, extracted, **kwargs):
```
patrick91: This fixture feels a lot like magic, maybe we should add some documentation around it to make it clear what it is doing :)

backend/submissions/models.py
marcoacierno: Do you think we should move this logic to the `Form`?

backend/api/forms.py
marcoacierno: Have a look now :)

backend/tests/conferences/factories.py
marcoacierno: Have a look now :)

backend/submissions/models.py
patrick91: I'd keep here for now, can't think of any reasons why we should move it to the forms :)

backend/api/forms.py
patrick91: can't see the docstring :P

patrick91: odd, had to refresh the page twice, now I see it :)

backend/conferences/models.py
marcoacierno: Added

backend/api/forms.py
marcoacierno: Let me know if it's clear enough now

PR #446, opened 2019-01-23
Schedule

marcoacierno:
- [ ] Schedule editor


marcoacierno:
@patrick91 What kind of features do you think we need for the schedule?

backend/conferences/types.py
```diff
@@ -58,12 +64,36 @@ class ConferenceType(DjangoObjectType):
     topics = graphene.NonNull(graphene.List(graphene.NonNull(TopicType)))
     languages = graphene.NonNull(graphene.List(graphene.NonNull(LanguageType)))
     durations = graphene.NonNull(graphene.List(graphene.NonNull(DurationType)))
+    schedule = graphene.NonNull(
+        graphene.List(graphene.NonNull(ModelScheduleItemType)),
+        date=graphene.String(),
```
patrick91: I think it might be better to use `graphene.types.datetime.Date` so we won't need to parse the date in the resolver :)

marcoacierno: Done

marcoacierno:
I added the `Room` model because I think it makes more sense than using the `topic` to decide the schedule room

backend/conferences/types.py
```diff
@@ -58,12 +64,30 @@ class ConferenceType(DjangoObjectType):
...
+        graphene.List(graphene.NonNull(ModelScheduleItemType)),
+        date=graphene.Date()
+    )
 
     timezone = graphene.String()
 
     def resolve_timezone(self, info):
         return str(self.timezone)
 
+    def resolve_schedule(self, info, date=None, room=None):
```
patrick91: are you using room? I don't think we need that filtering for now anyway :)

backend/tests/schedule/factories.py
```diff
@@ -0,0 +1,55 @@
...
+    def _create(cls, model_class, *args, **kwargs):
+        _type = kwargs.get('type', None)
+
+        if _type == ScheduleItem.TYPES.custom:
+            kwargs.pop('submission', None)
+
+        return super()._create(model_class, *args, **kwargs)
+
+    @factory.post_generation
+    def rooms(self, create, extracted, **kwargs):
```
patrick91: can you add a docstring here? not sure what it does :)

backend/conferences/types.py
marcoacierno: I think I forgot to remove it

marcoacierno: Done

backend/tests/schedule/factories.py
marcoacierno: Let me know if it's clear now

patrick91: 👍

PR #557, opened 2019-05-07
Refactor code to fix pylint warnings

marcoacierno:
I fixed all the issues that `pylint` reported.

The changes I did to the configuration:

- Increased `max-parents` to 15, we cannot do much here, sometimes you can reach that limit even with a simple django view, so I increased it
- Disabled `too-few-public-methods` because we have the graphql types that do not have any public method
- Disabled `no-member` because pylint is not able to understand Grahene well (so it thinks that `self.[field]` is `graphene.NonNull` instead of the model instance. Would love to fix this problem without disabling this rule as it's actually very useful

### DIFF
--- a/backend/api/mutations.py
+++ b/backend/api/mutations.py
@@ -15,8 +15,8 @@ class ContextAwareDjangoModelFormMutation(mutation.DjangoModelFormMutation):
         abstract = True
 
     @classmethod
-    def get_form_kwargs(cls, root, info, **input):
-        kwargs = super().get_form_kwargs(root, info, **input)
+    def get_form_kwargs(cls, root, info, **input_):
+        kwargs = super().get_form_kwargs(root, info, **input_)
         kwargs['context'] = info.context
         return kwargs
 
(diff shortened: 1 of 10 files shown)


marcoacierno:
Ugh. 
The upload to `codacy` is failing making the checks fail, this is not the first time and I don't even think we use `codacy` enough to justify the random fails. We might want to remove this service?

```requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://api.codacy.com/2.0/coverage/3a610c741e9345131d4cf68f959525f2d1a0285b/python```

File: backend/api/mutations.py [hunk 1]
```diff
@@ -15,8 +15,8 @@ class Meta:
         abstract = True
 
     @classmethod
-    def get_form_kwargs(cls, root, info, **input):
-        kwargs = super().get_form_kwargs(root, info, **input)
+    def get_form_kwargs(cls, root, info, **_input):
```
patrick91: wouldn't this be `input_`? :)

marcoacierno: done!

PR #566, opened 2019-05-11
Orders and Stripe integration

marcoacierno:
todo


marcoacierno:
@patrick91 it would be great to have an initial "WIP" review, just to see if you like the direction this PR is going :)

backend/payments/providers/provider.py
```diff
@@ -0,0 +1,14 @@
+from abc import ABC, abstractmethod
+
+
+class PaymentProvider(ABC):
```
patrick91: Love this

backend/payments/providers/stripe/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if payment_intent_id and payment_method_id:
+            raise ValueError(_('You cannot specify both payment_method_id and payment_intent_id'))
+
+        try:
+            if payment_method_id:
+                intent = stripe.PaymentIntent.create(
+                    payment_method=payment_method_id,
+                    amount=to_cents(order.amount),
+                    currency='eur', # TODO: make currency dynamic?
```
patrick91: could it be an attribute on the conference?

backend/payments/providers/stripe/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+                intent = stripe.PaymentIntent.confirm(payment_intent_id)
+            else:
+                raise ValueError(_("You need to specify at least a payment_method_id or a payment_intent_id"))
+        except CardError as e:
+            body = e.json_body
+            err  = body.get('error', {})
+
+            raise PaymentError(message=_(err.get('message')))
+        except RateLimitError as e:
+            raise PaymentError(message=_('Please try again in a few hours'))
```
patrick91: I was thinking, maybe we should return error codes and have the messages in the frontend, so we don't mix presentation and logic (kinda :)), it should make also things easier to translate in future (as we have only one place with strings)

backend/payments/providers/stripe/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+        except AuthenticationError as e:
+            raise PaymentError(message=_('Something went wrong on our side, please try again'))
+
+        if intent.amount != to_cents(order.amount):
```
patrick91: Oh interesting, when would this happen? (it's good to have the check, just curious)

backend/payments/providers/stripe/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+            # TODO: Better exception
+            raise ValueError(_('Payment intent amount and cart amount different'))
+
+        if intent.status == 'requires_action' and intent.next_action.type == 'use_stripe_sdk':
```
patrick91: we could move this to a function, so it is also easy to test with mocks :)

backend/payments/providers/utils.py
```diff
@@ -0,0 +1,11 @@
+from decimal import Decimal
+
+
+def to_cents(amount):
+    # I'm a bit scared here that we might call to_cents on a value
+    # already in cents causing issues.
+    # What can we do to make sure it does not happen?
```
patrick91: we could have a wrapping class for prices, instead of using decimal :)

backend/payments/types.py
```diff
@@ -0,0 +1,18 @@
+from graphene import Int, ID, NonNull, InputObjectType, ObjectType, String, List
+
+from conferences.types import TicketType
+from orders.types import OrderType
+
+
+class CartItem(InputObjectType):
+    id = NonNull(ID)
```
patrick91: Thanks for adding NonNulls :D

backend/pycon/settings/test.py
```diff
@@ -1,4 +1,9 @@
+from django.core.exceptions import ImproperlyConfigured
+
 from .base import *  # noqa
 
 
 SECRET_KEY = 'this-key-should-only-be-used-for-tests'
+
+if not STRIPE_SECRET_KEY.startswith('sk_test_'):
```
patrick91: brilliant!

backend/tests/graphene_form/test_converter.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def test_convert_boolean():
+    nonnull_boolean = convert_form_field(BooleanField())
+
+    assert isinstance(nonnull_boolean, Boolean)
+    assert nonnull_boolean.kwargs['description'] == ''
+    assert nonnull_boolean.kwargs['required'] is True
+
+    nonnull_boolean_with_desc = convert_form_field(BooleanField(help_text='eeeeeeeeeeeeeeewwwwwwww'))
```
patrick91: lol

backend/graphene_form/forms.py
```diff
@@ -0,0 +1,15 @@
...
+
+class FormWithContext(forms.Form):
+    def __init__(self, *args, **kwargs):  # pylint: disable=return-in-init
+        self._context = kwargs.pop('context', None)
+        return super().__init__(*args, **kwargs)
+
+    @property
+    def context(self):
+        if not self._context:
+            raise ValueError('Make sure you pass the context when instancing the Form')
```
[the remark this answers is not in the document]
marcoacierno: I wanted to keep `FormContext` flexible enough to be used even without context. I know it might not make sense when the name of the class has `Context` in it, but 😄

backend/graphene_form/converter.py
```diff
@@ -0,0 +1,80 @@
+from django import forms
```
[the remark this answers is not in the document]
marcoacierno: Actually it's just a copy and paste from graphene, I just wanted the module to be separate from the current implementation

backend/payments/forms.py
```diff
@@ -0,0 +1,141 @@
...
+        to_field_name='code',
+        required=True
+    )
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        conference = cleaned_data.get('conference', None)
+
+        if not conference:
```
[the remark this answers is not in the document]
marcoacierno: Clean gets called anyways, even if conference does not exists (and will fail validation)

backend/payments/providers/stripe/__init__.py
marcoacierno: Select less expensive items and try to retrieve the payment intent, then send a request to the backend with the `items` you really want and pass the intent id you got with the less expensive amount

marcoacierno:
I pushed the new code that uses the Webhook flow (and so the automatic payments flow) so that the 3d verification step is all managed by Stripe instead of us.

This change should also allow us to easily implement bank payments via Stripe if we wish to do so.

I know it's quite a huge change @patrick91, but do you have time to have a look and review? :)

Thank you! :)

PR #626, opened 2019-07-15
Add fannypack

fedriz:


frontend/src/components/button/button.mdx
```diff
@@ -3,15 +3,35 @@ name: Button
 ---
 
 import { Playground, Props } from "docz";
+// import { Button } from "fannypack";
```
patrick91: delete :D

frontend/src/components/button/button.mdx
```diff
@@ -3,15 +3,35 @@ name: Button
...
 import { Button } from "./";
+import { ThemeProvider } from "fannypack";
+import { theme } from "../../config/fannypack-config.ts";
```
patrick91: why .ts? :)

frontend/src/config/fannypack-config.ts
```diff
@@ -0,0 +1,36 @@
+import { css } from "fannypack";
+
+export const theme = {
+  // global: {
```
patrick91: delete :D delete :D delete :D

frontend/src/components/button/button.mdx
patrick91: let's add some aliases for this ;)

PR #632, opened 2019-07-15
Ticket Questions

greenkey:
Goal
- #633: Allow the users to answers questions when purchasing a ticket. 

Done
- created the model and the admin views for the questions, a TicketQuestion can be added to a TicketFare (M2M)
- created the model and the admin views (read-only) for the UserAnswers, they are basically the intersection table beween the Ticket and the TicketQuestionChoice (the possible answers for a Question)


marcoacierno:
The code looks good, but I think you are not running `pre-commit` because some files do not seem formatted

marcoacierno:
I'm going to close this one and review only #672, as #672 is basically this PR + the open question stuff.

PR #644, opened 2019-07-17
Make responsive styles utils more generic

patrick91:


frontend/src/components/columns/index.tsx
```diff
@@ -1,58 +1,80 @@
...
+  marginLeft?: ResponsiveValue;
+  marginRight?: ResponsiveValue;
+  paddingTop?: ResponsiveValue;
+  paddingBottom?: ResponsiveValue;
+  paddingLeft?: ResponsiveValue;
+  paddingRight?: ResponsiveValue;
+};
+
+export const responsiveSpacing = (props: ResponsiveProps) => css`
+  ${props.marginTop && responsiveStyle("marginTop", props.marginTop)}
```
fedriz: @patrick91 probably we can use the prop property name instead of the hardcoded name (eg: "marginTop")

PR #660, opened 2019-07-18
Feature/custom grid system

fedriz:


frontend/src/components/column/const.ts
```diff
@@ -0,0 +1,11 @@
+export const ALLOWED_RESPONSIVE_COLUMN_PROPS = [
+  "colWidth",
```
patrick91: can we change this to columnWidth? just to make it consistent with the other ColumnWidth* vars/types

frontend/src/components/column/index.tsx
```diff
@@ -1,13 +1,81 @@
 import React from "react";
 
-import { Column, LocalColumnProps } from "fannypack";
-import styled from "styled-components";
-import { ResponsiveProps, responsiveSpacing } from "../columns";
+import styled, { css } from "styled-components";
+import { customTheme } from "../../config/theme";
+import { ALLOWED_RESPONSIVE_COLUMN_PROPS } from "./const";
+import { camelToKebabCase } from "./utils";
```
patrick91: <3

frontend/src/components/column/index.tsx
```diff
@@ -1,13 +1,81 @@
...
+          }
+        `;
+      })}`;
+    });
+
+  return assembledStyle;
+};
+
+type ColumnSpacingType = {
+  colWidth?: ColumnWidthValuesType;
```
patrick91:
```suggestion
  columnWidth?: ColumnWidthValuesType;
```

PR #662, opened 2019-07-18
created blog app

schiism:


marcoacierno:
Looks good 🔥Can you run pre-commit so that the files are formatted?

backend/blog/models.py
```diff
@@ -0,0 +1,34 @@
...
+from django.utils import timezone
+from model_utils.models import TimeStampedModel
+
+class PostManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().filter(published__lte=timezone.now())
+
+
+class Post(TimeStampedModel):
+    author = models.ForeignKey('users.User', on_delete=models.CASCADE)
```
marcoacierno: I think I would use `models.SET_NULL` or `models.PROTECT`? So that we do not lose blog posts if we delete an admin account (I doubt this will ever happen however) What do you think?

PR #663, opened 2019-07-18
#35 Importer

### LINKED ISSUE #35: Importer
we need a way to import data from the old site to this one

greenkey:
Created the command `import_old_db` that takes the following parameters:
- `--db` to specify the full path of the source SQLite db file, default = `p3.db` (in the django root folder)
- `--entities` to specify which entities to import (see below)
- `--overwrite` some entities support this parameter to overwrite them while importing

The entities supported now are the following:
- `user`: imports User
- `submission_type`: SubmissionType
- `topic`: Topic
- `conference`: Conference and Deadline
- `room`: Room
- `submission`: 
- `schedule_item`: ScheduleItem
- `ticket`: TicketFare, Order, OrderItem, Ticket

Closes #35

### DIFF
--- a/backend/importer/management/commands/import_old_db.py
+++ b/backend/importer/management/commands/import_old_db.py
@@ -288,19 +288,23 @@ class Command(BaseCommand):
             code: submission_types[name]
             for code, name in TALK_TYPES.items()
         }
+        users_by_email = {
+            user.email: user.id
+            for user in User.objects.all()
+        }
 
         talk_list = list(self.c.execute(
             """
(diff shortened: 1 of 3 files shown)


patrick91:
I'm going to have a look at this soon, I think we should also find a way to show the data on the frontend so we can double check them in the ui as well (and it will help us in finishing the missing pages) :)

patrick91:
This is pretty cool! Tested locally and works fine :) I'll merge it, maybe we might need to updates in future, but it is great work so far! ✨

PR #669, opened 2019-07-19
Feature/graphs

estyxx:


patrick91:
@estyxx thanks for this! can we add a screenshot in the PR and maybe also a way for someone to test this? I want to send this to @ernestoarbitrio so he can have a look as he's the expert on bokeh :D

estyxx:
Here is the screenshot!
Screenshot 2019-07-23 at 23 33 56

I think maybe I could test the queries that get the data?

marcoacierno:
@estyxx can you rebase while we wait for @ernestoarbitrio feedback?

estyxx:
Hi, 
I just add a couple of test for that PR, but I still didn't reach 100% of coverage... I don't know how to test the functions building the graph really...

marcoacierno:
We are removing everything related to the tickets in the backend in another PR so I think we can close this PR?

estyxx:
But we need some statistics info... Maybe we need to change the query for getting the data and extract data from prefix?

estyxx:
Maybe we should do a better feature
/pr to include all @ernestoarbitrio statistics?? A separate page (I don't know if we could include also jupiter notebook...?)

PR #672, opened 2019-07-20
Add ticket questions support

greenkey:
Done:
- change answer model to support open answers
- make TicketQuestionChoices.answer a text field, the value will be given by the caller, picking one of the coiches values
- add a bit of validation on answers so that only allowed answers are saved
- add the missing migrations


backend/tickets/models.py
```diff
@@ -21,3 +24,62 @@ class Ticket(TimeStampedModel):
...
+        max_length=256,
+        null=True,
+    )
+
+    def save(self, *args, **kwargs):
+        self.full_clean()
+        super(UserAnswer, self).save(*args, **kwargs)
+
+    def clean(self):
+        # Don't allow draft entries to have a pub_date.
```
marcoacierno: Is this comment a copy and paste typo?

backend/tickets/models.py
```diff
@@ -21,3 +24,62 @@ class Ticket(TimeStampedModel):
...
+        if self.question.question_type == QUESTION_TYPE_CHOICE:
+            if self.answer not in self.question.choices.all().values_list('choice', flat=True):
```
marcoacierno: Hmmm, I wonder if we can somehow make answer a reference to the question choice so to avoid this check?

marcoacierno: well, make it a bit different (we still need to check that the answer is for the correct question/anything else), but still  better than a free text?

marcoacierno:
I feel like this PR is now complete @patrick91, we added the GraphQL mutations to answer a question and the changes we discussed (add `is_required` to specify a question can be answered blank or not.) 

The only thing missing is how to get the user's answer in GraphQL.

My initial idea was to basically return the answer of each question in the `tickets`

```graphql
  me {
    tickets(conference: "cc") {
      ticketFare {
        name
      }
      answers {
	answer
        question {
          text
        }
      }
    }
  }
```

which means that you have to somehow map the `answers` (which might not be all the questions) and the `questions` from the `ticketFare`

You might do a query like this:

```graphql
  me {
    tickets(conference: "cc") {
      ticketFare {
        name
        questions {
          isRequired
          question {
            text
          }
        }
      }
      answers {
	answer
        question {
          text
        }
      }
    }
  }
```

and the answer will look like this:

```json
    "me": {
      "tickets": [
        {
          "ticketFare": {
            "name": "aa",
            "questions": [
              {
                "isRequired": true,
                "question": {
                  "text": "Tagline"
                }
              },
              {
                "isRequired": false,
                "question": {
                  "text": "t-shirt size"
                }
              }
            ]
          },
          "answers": [
            {
              "answer": "👀",
              "question": {
                "text": "Tagline"
              }
            }
          ]
        }
      ]
    }
```

do you have any idea?

patrick91:
I'd something like:

```gql
  me {
    tickets(conference: "cc") {
      ticketFare {
        name
        questions {
          isRequired
          question {
            text
          }
          answer # so the frontend doesn't need to map anything
        }
      }
  }
```

backend/conferences/types.py
```diff
@@ -8,6 +8,9 @@
 from schedule.types import Room, ScheduleItem
 from submissions.types import Submission, SubmissionType
 
+if False:
```
patrick91: why do we need this? if it is for the type checking we can use TYPE_CHECKING: https://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles

marcoacierno: fixed!

patrick91:
Looks good! Let’s get this merged!

PR #788, opened 2019-08-23
Feature/users profile Edit

estyxx:
Before continue developing is better to have a look I think... :)

Here is a list of thing to-do/question to compete: 
- [x] **FE** fix data input


backend/.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -0,0 +1,20 @@
+<component name="InspectionProjectProfileManager">
```
marcoacierno: Can you add to gitignore the .idea specific files?

backend/users/forms.py
```diff
@@ -58,3 +63,57 @@ def save(self):
         user = authenticate(email=email, password=password)
         login(self.context, user)
         return user
+
+
+class UpdateForm(ContextAwareModelForm):
+    date_birth = CharField(required=False)
+    # TODO: Why if I tell you required=FALSE, the mutation fail if I don't put
+    # Bool! (instead of 'Bool')?
+    open_to_recruiting = BooleanField(required=False)
```
marcoacierno: I think if you check the mapping it says that BooleanField is always mapped to a required parameter

backend/users/forms.py
```diff
@@ -58,3 +63,57 @@ def save(self):
...
+    open_to_newsletter = BooleanField(required=False)
+    country = CharField(required=False)
+
+    def clean(self):
+
+        if not self.context.user.is_authenticated:
+            raise ValidationError(_("Must authenticate to update User information"))
```
marcoacierno: Don't we have a `permission_classes` to avoid to insert this check everywhere? I think if you have a look at `SendSubmission` you should find an use of that

backend/users/forms.py
```diff
@@ -58,3 +63,57 @@ def save(self):
...
+
+        return super().clean()
+
+    def save(self, commit=True):
+        super().save(commit=commit)
+        return User.objects.get(id=self.context.user.id)
```
marcoacierno: I think `self.instance` should already be up to date with the changes? Maybe we can avoid a new query to get the updated?

backend/users/models.py
```diff
@@ -1,15 +1,81 @@
...
+from django.core import exceptions
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
+from model_utils import Choices
+from pycountry import countries
 
 from .managers import UserManager
 
+COUNTRIES = [{"code": country.alpha_2, "name": country.name} for country in countries]
+EU_COUNTRIES = (
```
marcoacierno: We have to decide if we want russia to be in the EU_COUNTRIES or not

backend/users/models.py
```diff
@@ -1,15 +1,81 @@
...
+
+[
+    {"code": country.alpha_2, "name": country.name}
+    for country in countries
+    if country.alpha_2 in EU_COUNTRIES
+]
+
 
 class User(AbstractBaseUser, PermissionsMixin):
+
```
marcoacierno: Why the blank space here?

backend/users/models.py
```diff
@@ -1,15 +1,81 @@
...
+    vat_number = models.CharField(_("VAT number"), max_length=22, blank=True)
+    phone_number = models.CharField(_("phone number"), max_length=20, blank=True)
+
+    # electronic invoicing ita
+    recipient_code = models.CharField(_("recipient code"), max_length=7, blank=True)
+    pec_address = models.EmailField(_("PEC"), blank=True)
+
+    address = models.TextField(_("address"), blank=True)
+    country = models.CharField(
+        choices=COUNTRIES, max_length=50, verbose_name=_("country"), null=True
```
marcoacierno: Maybe `blank=True` instead of null? As it seems more fitting to leave it empty (adding default='')

backend/users/types.py
```diff
@@ -1,21 +1,58 @@
...
+    date_birth: str
+    business_name: Optional[str]
+    fiscal_code: Optional[str]
+    vat_number: Optional[str]
+    recipient_code: Optional[str]
+    pec_address: Optional[str]
+    address: Optional[str]
+    country: str
+    phone_number: Optional[str]
+    image: Optional[Image]
```
marcoacierno: Image is defined two times here, and below with a resolver

backend/users/models.py
```diff
@@ -23,3 +89,106 @@ class User(AbstractBaseUser, PermissionsMixin):
...
+            return
+
+        if not self.vat_number:
+            raise exceptions.ValidationError(
+                {"vat_number": _("Missing VAT Number in your user profile.")}
+            )
+
+    def clean_fiscal_code(self):
+        msg = _("Fiscal Code is invalid")
+        if not len(self.fiscal_code) == 16:
```
marcoacierno: I am not a super expert of the italian fiscal code, but if we use the User model for companies as well the length of the code might be 11

backend/.idea/inspectionProfiles/Project_Default.xml
estyxx: yes of course

backend/users/models.py
estyxx: Fiscal Code is different from the VAT maybe you are confusing the two fields? 
Fiscal Code is only for people or freelancer I think...

backend/.idea/inspectionProfiles/Project_Default.xml
estyxx: added

backend/users/models.py
estyxx: Maybe as EU countries we should consider only countries with the euro money??

backend/users/models.py
estyxx: oks

backend/users/forms.py
estyxx: Ah oks, i removed the todo...

backend/users/models.py
estyxx: This method is specific for italian fiscal code ("SSSNNNYYMDDZZZZX"). If there are some roles for VAT or companies code  I have to develop another method to check that code... 
Maybe I can put a docstring with these notes...

backend/users/types.py
estyxx: Mmm this could be a bug of strawberry? If leave only the resolver, I got an error when it tries to instantiate the `MeUserType` in `BaseUserMutation.transform`:

'__init__() got an unexpected keyword argument \'image\''

backend/users/forms.py
estyxx: Yup, of course!

backend/users/models.py
estyxx: As we decided to remove any company field, for now, I resolve this.

backend/users/forms.py
estyxx: fixed

backend/users/types.py
marcoacierno: Looks like a bug of Strawberry indeed, I will have a look

backend/.idea/inspectionProfiles/Project_Default.xml
marcoacierno: Looks like the files are still pushed in the commit?

estyxx: Now they should be removed...

backend/users/models.py
estyxx: EU_COUNTRIES are those in CEE, so [this list](https://en.wikipedia.org/wiki/European_Union#Member_states). So they are right... :)

backend/pycon/settings/base.py
```diff
@@ -146,3 +148,24 @@
 
 STATICFILES_STORAGE = "whitenoise.storage.CompressedManifestStaticFilesStorage"
 MAPBOX_PUBLIC_API_KEY = env("MAPBOX_PUBLIC_API_KEY", default="")
+
+USE_AWS_S3 = env("USE_AWS_S3", default=False) == "True"
```
marcoacierno: Once you are done testing can you move the settings back to prod? If we need to test again we can set `DJANGO_MODULE_SETTINGS` to `pycon.settings.prod` to change the settings that `runserver` loads

backend/pycon/settings/base.py
```diff
@@ -146,3 +148,24 @@
...
+    # aws settings
+    AWS_ACCESS_KEY_ID = env("AWS_ACCESS_KEY_ID")
+    AWS_SECRET_ACCESS_KEY = env("AWS_SECRET_ACCESS_KEY")
+    AWS_STORAGE_BUCKET_NAME = env("AWS_MEDIA_BUCKET")
+    AWS_S3_OBJECT_PARAMETERS = {"CacheControl": "max-age=86400"}
+    AWS_DEFAULT_ACL = "public-read"
+    AWS_QUERYSTRING_AUTH = False
+    AWS_S3_CUSTOM_DOMAIN = f"{AWS_STORAGE_BUCKET_NAME}.s3.amazonaws.com"
+
+    # s3 static settings
```
marcoacierno: Do we need all those settings? @patrick91

backend/pycon/storage_backends.py
```diff
@@ -0,0 +1,12 @@
+from storages.backends.s3boto3 import S3Boto3Storage
+
+
+class StaticStorage(S3Boto3Storage):
+    location = "static"
```
marcoacierno: Hmm, do we want to use the same bucket for the static files? I would go for a different bucket and not a different path, bucket are free and I feel safer with media and static files separated.

What do you think @patrick91?

backend/tests/upload/test_model.py
```diff
@@ -0,0 +1,58 @@
...
+def file_sample():
+
+    path = "testfile.txt"
+    f = open(path, "w")
+    f.write("Hello World")
+    f.close()
+
+    post_data = {"file": open(path, "rb")}
+
+    yield (path, post_data)
```
marcoacierno: This looks like a good solution to clean up the files once you are done with them, I usually just move the tests in a temporary media folder and add a pytest post hook to clean up the folder once all the tests are run. This one seems like an interesting solution

backend/upload/views.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+def file_upload(request):
+    if request.method != "POST":
+        return HttpResponseForbidden()
+
+    file = request.FILES["file"]
+    upload = Upload(file=file)
+    upload.save()
+    if not settings.USE_AWS_S3:
```
marcoacierno: Why the difference?

backend/users/forms.py
```diff
@@ -58,3 +64,41 @@ def save(self):
...
+
+    class Meta:
+        model = User
+        fields = ("first_name", "last_name", "gender")
+
+
+class UpdateImageForm(FormWithContext):
```
marcoacierno: Why the image has a different form mutation?

Maybe you can check that if the `image` is specified in the form's payload you can run the code you are now running in the other form?

Just an idea

backend/upload/views.py
estyxx: Because with S3 storage `upload.file.url` gives me the complete url (ex: `"https://pycon-static.s3.amazonaws.com/media/tmp/sample.png"`) and it's not fine when I have to retrieve in the `UpdateImageForm` with `default_storage.open(url)` because it wants the relative path, this: `'tmp/sample.png'` otherwise the file will not be found! 

And when the file is stored in the file system I need the absolute path because `upload.file.url` gives me this: `'/media/tmp/sample_sHbAJpO.png'` with the `/media` directory included... Really, the file is in `MEDIA_ROOT` + `tmp/sample_sHbAJpO.png`.

backend/users/forms.py
estyxx: Yes, true. I thought to divide because usually in the FE the update of the image is done separately, but in the BE is only more code...

estyxx: done in [e4d51ce](https://github.com/pythonitalia/pycon/pull/788/commits/e4d51ce74487f20cc4089e100cb2d06cb8b6e864)

backend/upload/models.py
```diff
@@ -0,0 +1,13 @@
+from django.db import models
+from django.utils.translation import ugettext_lazy as _
+
+
+class File(models.Model):
+    file = models.FileField(_("file"), upload_to="tmp", blank=False, null=False)
+    date_upload = models.DateTimeField(_("date upload"), auto_now_add=True)
+
```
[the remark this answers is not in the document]
estyxx: @patrick91 @marcoacierno should we track who upload the file? [maybe optionally because one can upload a file without being logged...]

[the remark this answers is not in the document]
marcoacierno: do we still need this model?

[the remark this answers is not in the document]
estyxx: Do we have another (better) option?? I use it to save the file in views.py... (p. s. was rename to Upload..)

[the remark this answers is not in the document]
estyxx: Maybe with `graphql-upload` we can upload the files directly? [See link](https://levelup.gitconnected.com/how-to-add-file-upload-to-your-graphql-api-34d51e341f38) I should have a try... :)

frontend/src/app/profile-edit/index.tsx
```diff
@@ -0,0 +1,342 @@
...
+                          },
+                        })}
+                      isHorizontal={true}
+                      a11yId="gender"
+                      label={
+                        <FormattedMessage id="profile.gender">
+                          {msg => <b>{msg}</b>}
+                        </FormattedMessage>
+                      }
+                      options={[
```
estyxx: These options are hardcoded, of course, they are pretty simple and should not change... but at least we should translate, can I take values from `locale.index.tsx` dictionary? [I can't use FormattedMessage component... ]
Is ugly?? Is better put somewhere a resolver in the BE?

frontend/src/app/profile-edit/index.tsx
```diff
@@ -0,0 +1,342 @@
...
+                          {msg => <b>{msg}</b>}
+                        </FormattedMessage>
+                      }
+                      onChange={hangleUserChange}
+                    />
+                  </InputWrapper>
+
+                  <InputWrapper text={errorsFields.gender} isRequired={true}>
+                    <RadioGroupField
+                      {...radio("gender",
```
estyxx: The components `radio`, `select`, `raw` of react-use-form-state they wrapped the field with a different id, 
if I use hangleUserChange it save the change in a different random field.. so I re-do the update here manually... 
I don't like very much...

estyxx:
I think this PR start to be complete... I think you @patrick91 could have a look :)
I only have to understand why "gender" doesn't set when the MyProfileQuery fetches the data... maybe is something related to the radio component of /react-use-form-state... apart from that the page is working... :)

backend/users/tests/test_update_user.py
```diff
@@ -0,0 +1,122 @@
...
+        gender="male",
+        open_to_recruiting=True,
+        open_to_newsletter=True,
+        date_birth=datetime.datetime.strptime("1940-10-09", "%Y-%m-%d"),
+        country="AT",
+    )
+    graphql_client.force_login(user)
+    resp, variables = _update_user(graphql_client, user)
+    resp, variables = _update_user(graphql_client, user)
+    print(resp)
```
marcoacierno: I think this one is there for debug right? Also the duplicate the _update_user call

frontend/src/templates/app.tsx
```diff
@@ -17,6 +18,7 @@ export default () => (
   >
     <Router>
       <PrivateRoute path="/:lang/profile" component={ProfileApp} />
+      <PrivateRoute path="/:lang/profile-edit" component={EditProfileApp} />
```
marcoacierno: Maybe `profile/edit`? (it's not important to be honest)

PR #823, opened 2019-09-01
Add meta information about conference

patrick91:
This PRs add info like introduction and generic copy text, like the screenshot :)

image


backend/conferences/models/conference.py
```diff
@@ -32,6 +32,7 @@ class Conference(TimeFramedModel, TimeStampedModel):
         _("longitude"), max_digits=9, decimal_places=6, blank=True, null=True
     )
     map_link = models.URLField(_("map link"), blank=True)
+    introduction = models.TextField(_("introduction"), blank=False)
```
marcoacierno: Are we sure we want to make the introduction as required?

backend/conferences/types.py
```diff
@@ -130,6 +131,14 @@ def sponsors_by_level(self, info) -> List[SponsorsByLevel]:
 
         return [SponsorsByLevel(level, list(sponsors)) for level, sponsors in by_level]
 
+    @strawberry.field
+    def copy(self, info, key: str) -> Optional[str]:
+        from cms.models import GenericCopy
```
marcoacierno: Any reason for the import here?

backend/conferences/models/conference.py
patrick91: yes, it is required on the frontend :)

backend/conferences/types.py
patrick91: need to double check :)

patrick91: fixed :)

PR #885, opened 2019-09-10
Use GraphQL view from strawberry

patrick91:


marcoacierno:
This looks good! Do we want to wait to fix the codecov path before merging?

PR #909, opened 2019-09-14
Move apps api components into api

marcoacierno:
Moved everything that's api related into the `api` app.

Should we move the tests into `tests/api` too?


patrick91:
@marcoacierno yes, let's move the tests as well :)

I'd move them into api/tests ;) let's try to to put tests near the code we are testing :)

marcoacierno:
Do you want to move all the tests in their app directory?

patrick91:
@marcoacierno yes, but can be done in another pr as well :)

marcoacierno:
done! @patrick91 have a look :)

.github/workflows/backend-test.yml
```diff
@@ -22,12 +22,11 @@ jobs:
       - name: Install codecov
         working-directory: backend
         run: pip install codecov
-
       - name: pytest
         working-directory: backend
         run: |
             set +e
-            OUTPUT=$(poetry run py.test --cov-report xml --cov=. --cov-fail-under 100)
+            OUTPUT=$(poetry run pytest */tests/ --cov-report xml --cov=. --cov-fail-under 100)
```
patrick91: I think this could be `.`, as pyrest only gets files starting with `test_` :)

.github/workflows/backend-test.yml
```diff
@@ -48,9 +47,8 @@ jobs:
       - name: Install dependencies
         working-directory: backend
         run: poetry install
-
       - name: pytest
         working-directory: backend
-        run: poetry run py.test integration_tests/
+        run: poetry run pytest */integration_tests/
```
patrick91: uhm, we can either mark integration tests with something like @pytest.skip.if(INTEGRATION=False) or have them somewhere else :)

marcoacierno: Do you have any preference?

.github/workflows/backend-test.yml
marcoacierno: This way  integration tests are not run

.github/workflows/backend-test.yml
patrick91: something like this if it works ;) https://gist.github.com/billyshambrook/264624b039fb3caa9278

marcoacierno: have a look

patrick91: top! we can merge ;)

PR #965, opened 2019-09-26
Add invoke to make setup easier

patrick91:


estyxx:
Please when we will back to this PR please change `.env.sample` with the default values 

```
API_URL=http://127.0.0.1:8000/graphql
CONFERENCE_CODE=pycon-demo
```
now is more similar to what we have in production... could be hard to identify the problem...
 
tnks :)

PR #1033, opened 2019-10-06
Improve docs: conference entity description

estyxx:
This It's just an introduction about the main entities of our backend... I took the original content from Notion and organized/integrate it...
Maybe, if @marcoacierno could integrate with orders/payment little description would be perfect :)


docs/team.md
```diff
@@ -0,0 +1,6 @@
+# Team
```
patrick91: let's remove this, don't think we need to show the team in the docs :D also you forgot yourself :P

estyxx: I mention only the team leaders...
Maybe not a whole page (I don't have too much experience with open source...) but should we mention some people, a section, "who you can ask if you have questions..."...

patrick91: Uhm, I feel like questions should be asked as issues, so anyone could reply ;) and we also have a record of the questions  :)

PR #1052, opened 2019-10-07
Add signup form

marcoacierno:


frontend/src/components/signup-form/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  if (loggedIn) {
+    return <Redirect to={profileUrl} noThrow={true} />;
+  }
+
+  const errorMessage =
+    data && data.register.__typename === "RegisterErrors"
+      ? data.register.nonFieldErrors.join(" ")
+      : error;
+
+  const getFieldError = (field: "validationEmail" | "validationPassword") =>
```
patrick91: we should move this to an utility in another PR :)

PR #1103, opened 2019-10-13
Update gatsby and fannypack

patrick91:


patrick91:
https://github.com/fannypackui/fannypack/issues/85

PR #1104, opened 2019-10-13
docker commands added to setup, DOCKER and DB vars added to .env.sample

madEng84:
Hi @patrick91, 
I think the best way to configure a postgres DB is using docker with a separated volume.
Suddenly you can see in .env.sample some vars used in `setup-db` invoke command (`DB_USER`, `DB_PASSWORD` etc.), but not used as vars in `DATABASE_URL` .env var (we cannot use an .env var as var for other .env vars)
I'm not satisfied, maybe we can find a better solution, but it can be considered a good start at least to be able to use a postgres DB


tasks.py
```diff
@@ -9,20 +9,79 @@ def setup_frontend(c):
...
+def setup_db(c):
+    print("Running DB setup...")
+
+    with c.cd("backend"):
+        try:
+            c.run("cat .env", hide=True)
+        except Failure as ex:
+            print("ERROR! Please copy backend/.env.sample in backend/.env and set your "
+                  ".env vars")
+            raise ex
```
patrick91: do we need this raise? :) might not be that user friend, we already show a useful message anyway 😊

patrick91: same for the other ones, maybe we could create a utility, something like:

```
run_or_fail(c, "docker -v", error_message="Docker missing...")
```

tasks.py
```diff
@@ -9,20 +9,79 @@ def setup_frontend(c):
...
+
+    print("... DB setup completed!")
+
+
 @task()
 def setup_backend(c):
     print("Running backend setup...")
 
     with c.cd("backend"):
+        c.run("pip install poetry", hide=True, warn=True)
```
patrick91: uhm, I wouldn't do this, I would let the users know that poetry is missing and then let them install it they way they prefer :)

tasks.py
madEng84: Good idea ;D

tasks.py
madEng84: Yes, you're right

tasks.py
```diff
@@ -1,76 +1,77 @@
...
+def tools_check(c, tools=[]):
+    missing_tools = []
+    for tool in tools:
+        if c.run(f"which {tool}", hide=True, warn=True).exited != 0:
+            missing_tools.append(f"{tool}")
+
+    if len(missing_tools) > 0:
+        print("These tools are required : {}".format(", ".join(missing_tools)))
+        print("Please install them")
+        exit()
```
madEng84: @patrick91 I used the approach you recommended to me, but I envisaged the possibility of passing the tools to be checked by the task so that it could be used as a prerequisite for other tasks.

I hope you like it :D

tasks.py
```diff
@@ -1,76 +1,77 @@
...
+
+
+def run_or_fail(c, cmd, fail_message="Error", fail_callback=None):
+    try:
+        cmd(c)
+    except Failure as ex:
+        print(fail_message)
+        if fail_callback:
+            fail_callback(c)
+        exit()
```
madEng84: This approach is similar to the one you suggested to me, but I preferred to pass a function rather than a command to have the possibility to pass parameters to the function without having to add other input parameters
I have also foreseen the possibility of passing a callback in case of exception

tasks.py
```diff
@@ -1,28 +1,87 @@
...
+                                 "in backend/.env and set your .env vars")
+        with c.prefix(". .env"):
+            run_or_fail(c, lambda x: x.run("docker info", hide=True),
+                        fail_message="ERROR! Is Docker running?")
+            run_or_fail(c, lambda x: x.run("docker volume create $DOCKER_DB_VOLUME", hide=True),
+                        fail_message=f"ERROR! Docker volume '$DOCKER_DB_VOLUME' has not "
+                                     f"been created\n"
+                                     f"Here the results of command 'docker volume inspect $DOCKER_DB_VOLUME'",
+                        fail_callback=lambda x: x.run("docker volume inspect $DOCKER_DB_VOLUME"))
+            run_or_fail(c, lambda x: x.run("docker start $DOCKER_DB_CONTAINER ||"
```
marcoacierno: Maybe it makes sense to consider using docker-compose instead of the docker run command?

marcoacierno: So we can handle all the volumes etc stuff there, instead of running the various docker commands

tasks.py
marcoacierno: I like this tools check ✨

tasks.py
madEng84: Yes @marcoacierno, i'm already working on it. I want to serve only db with docker-compose, is it ok?

marcoacierno:
I also had an idea to create an `invoke` command to basically setup the initial required fields (which  for now I think it's only `.env` copying from `.env.sample`), what do you think?

tasks.py
marcoacierno: Hmm, it might definitely work as a start

Also I don't know if in the future we also want to move the backend on `docker-compose` and bind the local code when running with docker-compose.local.yml (for example). Or use docker-compose only for local deveploment as we have all the AWS stuff for production

Let me know if it's not clear what I'm thinking! :)

madEng84: Yes it's clear.
I agree with you

madEng84:
@marcoacierno do you want to make checks on personal .env vars?

marcoacierno:
My idea was to have a command like `invoke init-env` that prepares the project. Right now it's a simple `copy .env.sample to .env` but it can be expanded in the future

tasks.py
marcoacierno: The only issue I see is in case any tool, for some reason, do not work with `which` but I we can handle that in the future

tasks.py
```diff
@@ -1,28 +1,70 @@
...
 
 
-@task()
+@task(pre=[call(tools_check, ["docker", "docker-compose"])])
+def setup_db(c):
+    print("Running DB setup...")
+
+    with c.cd("backend"):
+        run_or_fail(c, lambda x: x.run("ls .env.sample", hide=False),
+                    fail_message="Are you in project root dir?")
```
marcoacierno: Don't we change the directory to `backend` manually using `c.cd`? Do we need that?

tasks.py
```diff
@@ -1,28 +1,70 @@
...
+        run_or_fail(c, lambda x: x.run("ls .env", hide=False),
+                    fail_message="ERROR! Have you copied backend/.env.sample "
+                                 "in backend/.env and set your .env vars?")
```
marcoacierno: Maybe it's useful to some util for preconditions? 

So this code can be:

```
[...]

check_env_precondition()
```

in case we need to reuse this same check somewhere else

marcoacierno:
I left like 1-2 comments but nothing serious, the `docker-compose` for postgres looks ok to me.

The reason the Github checks are failing is because you are using your own fork instead of a branch of the site, so the actions do not get pycon's private environment variables, making pretty much everything fail.

tasks.py
madEng84: 🤦‍♂
obviously we don't need it  😄

madEng84:
@marcoacierno I made the requested changes
I hope it's ok

marcoacierno:
Hi @madEng84, closing for now as we are changing a few things. Let's discuss on Discord or Slack what to do to our setup and then create a new PR

Thanks! :)

PR #1198, opened 2019-10-29
Add PrivateRoute to redirect non-authenticated user to login pa…

estyxx:
Some pages require that user be authenticated, if not we redirected to the login page. 
I created a PrivateRoute component that checks it and Redirect to login if is not authenticated otherwise it Redirect to the component.

### DIFF
--- a/frontend/src/app/private-route/private-route.tsx
+++ b/frontend/src/app/private-route/private-route.tsx
@@ -1,0 +1,15 @@
+import { useLoginState } from "../profile/hooks";
+import React, { Component } from "react"
+import { Redirect } from "@reach/router";
+
+
+export const PrivateRoute = ({component: Component, location, lang, ...rest}: any) => {
+  const [loggedIn, _] = useLoginState(false);
+  const loginUrl = `${lang}/login`;
+
+  if (!loggedIn && !location.pathname.includes("login")) {
+    return <Redirect to={loginUrl} noThrow={true} />;
+  }
+
+  return <Component {...rest} />
+};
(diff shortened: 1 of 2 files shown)


frontend/src/app/private-route/private-route.tsx
```diff
@@ -0,0 +1,15 @@
+import { useLoginState } from "../profile/hooks";
+import React, { Component } from "react"
+import { Redirect } from "@reach/router";
+
+
+export const PrivateRoute = ({component: Component, location, lang, ...rest}: any) => {
+  const [loggedIn, _] = useLoginState(false);
+  const loginUrl = `${lang}/login`;
+
+  if (!loggedIn && !location.pathname.includes("login")) {
```
marcoacierno: Why this `location.pathname.includes("login")` check?

estyxx: Because if you are already in the "login" page we don't need to redirect... 
[location.path includes `/{lang}/` so I checked _if includes_ login, not if is equal...]

patrick91: The loging route won't be a PrivateRoute no? :) it would never get here

estyxx: Ah, That's right! 
I copied the [documentation](https://www.gatsbyjs.org/tutorial/authentication-tutorial/#controlling-private-routes) 😅, but in our case, we don't need it :)

estyxx: fixed :)

PR #1222, opened 2019-11-02
Allow Submission to support multiple languages

estyxx:


backend/submissions/models.py
```diff
@@ -62,17 +62,6 @@ def clean(self):
...
-            self.language_id
-            and not self.conference.languages.filter(id=self.language_id).exists()
-        ):
-            raise exceptions.ValidationError(
-                {
-                    "language": _("%(language)s is not an allowed language")
-                    % {"language": str(self.language)}
-                }
-            )
-
```
marcoacierno: Any reason why we prefer the validation in the form instead of the model now? Do we want to move the other checks as well?

marcoacierno:
I left one comment but otherwise it looks ok to me! ✨

backend/submissions/models.py
estyxx: Yes, because becoming many-to-many I don't have the instance "languages" and I can not retrieve... I don't remember the exact message at the moment, Django says something about languages aren't ready jet during the model validation...

estyxx:
```
{ValueError}"<Submission: Player they evening war true magazine start. at Conference 1>" needs to have a value for field "id" before this many-to-many relationship can be used.
```
this is the error when you try to access to `self.languages` in the clean of the model Submission

marcoacierno:
@estyxx can you rebase so we can merge it? :)

marcoacierno:
Thank you! ✨

PR #1223, opened 2019-11-03
New design Footer

marcoacierno:


marcoacierno:
I hardcoded the social addresses because I'm not suuper sure where to put them, but at the same time I feel like it's ok to leave them in the frontend as they won't change? :)

PR #1244, opened 2019-11-07
Feature/add newsletter

estyxx:
Sorry, I redid this PR because there was something weird in the old... 

- [ ] Do subscription serverside in order to avoid spam
    - [ ] Remove `react-use-mailchimp` 
    - [ ] Write/send mutation to BE 
    - [ ] Use /react-google-recaptcha 
    - [ ] Use mailchimp3 for BE
    - [ ] Was addes a Url for store mailchimp link, maybe now we can store mc_api, mc_user, id list? 
    - [ ] Do we need a separate component for mailchimp?? mmm i don't think so...


patrick91:
@estyxx can you do this PR against this other PR? https://github.com/pythonitalia/pycon/pull/1219

PR #1274, opened 2019-11-17
Google social login

marcoacierno:


frontend/src/components/login-form/index.tsx
```diff
@@ -54,43 +55,81 @@ export const LoginForm: React.SFC<RouteComponentProps<{ lang: string }>> = ({
...
-        id="login-password"
-        {...password("password")}
-        required={true}
-        type="password"
-      />
-
-      <Button size="medium" palette="primary" isLoading={loading} type="submit">
-        <FormattedMessage id="login.loginButton" />
-      </Button>
+      {location && location.state && location.state.message && (
```
patrick91: nit: I think we can use the ? operator now: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

frontend/src/components/login-form/index.tsx
```diff
@@ -54,43 +55,81 @@ export const LoginForm: React.SFC<RouteComponentProps<{ lang: string }>> = ({
...
+          maxWidth: "container",
+          mx: "auto",
+          mt: 3,
+          gridTemplateColumns: [null, null, "1fr 1fr"],
+          gridColumnGap: 5,
+        }}
+      >
+        <Box
+          sx={{
+            maxWidth: "smallContainer",
```
patrick91: do we need a small container as well? :D

marcoacierno: Uhm, no actually! that was before I changed the parent to be a grid, nice catch thanks

frontend/src/components/login-form/index.tsx
marcoacierno: Changed!

backend/pycon/settings/base.py
```diff
@@ -145,7 +145,7 @@
 SOCIAL_AUTH_POSTGRES_JSONFIELD = True
 
 SOCIAL_AUTH_LOGIN_REDIRECT_URL = f"{FRONTEND_URL}/login/success/"
-SOCIAL_AUTH_LOGIN_ERROR_URL = f"{FRONTEND_URL}/login/error/"
+# SOCIAL_AUTH_LOGIN_ERROR_URL = f"{FRONTEND_URL}/login/"
```
patrick91: do we need this? :)

PR #1299, opened 2019-11-20
Feature Tag line

estyxx:
Something like [that](https://medium.com/@jerrylowm/build-a-tags-input-react-component-from-scratch-1524f02acb9a) for the frontend :)

The _Tag-Line_ will be used in the `CfpForm`,  the `SubmissionPage` and in `VotingPage` for filter submission maybe. Only in the `CfpForm` the user will see the `x` to remove the tag. 

### TO-DO
[ ] Make `x` button optional


backend/api/submissions/types.py
```diff
@@ -33,6 +39,7 @@ class Submission:
     duration: "Duration"
     audience_level: "AudienceLevel"
     languages: List["Language"]
+    tags: Optional[List[SubmissionTag]]
```
patrick91:
```suggestion
    tags: List[SubmissionTag]
```

The list will be empty by default no? No need to make it optional :)

backend/api/submissions/types.py
```diff
@@ -49,3 +56,7 @@ def my_vote(self, info) -> Optional[VoteType]:
     @strawberry.field
     def languages(self, info) -> List[Language]:
         return self.languages.all()
+
+    @strawberry.field
+    def tags(self, info) -> Optional[List[SubmissionTag]]:
```
patrick91:
```suggestion
    def tags(self, info) -> List[SubmissionTag]:
```

backend/api/tests/schema/test_submission.py
```diff
@@ -47,3 +50,28 @@ def test_returns_correct_submission(graphql_client, user, submission_factory):
...
+            }) {
+                __typename
+                ... on SubmissionTag{
+                    name
+                }
+                ... on SendTagErrors {
+                    nonFieldErrors
+                }
+            }
+        }
```
patrick91: why do we need a send tag mutation? can't we do that when we send/edit a submission?

patrick91: or is this to create new tags? I think we should allow only a few tags decided by us :)

Or if we want to allow any kind of tags I guess at that point we can use an array field instead of having a model for the tags :)

estyxx: No, I don't know what I was thinking... Maybe when in the frontend you hit the comma and appears the tag-button I think to do a mutation to save that tag initially... 
I removed it

Issue #1090, opened 2019-10-10
Add mutation to subscribe user to newsletter


patrick91:
We need this for the launch, we are using mailchimp, I can setup api keys when needed :)

estyxx:
What do you mean with this?
In the user model, we have `openToNewsletter` field, but you mean *any user* I think... anyone, also who is not registered? Maybe we have to put in a separate model, with an FK to the user model if the newsletter-user is already registered (the same email) or if he will...? 

If we choose to have a dedicated model for this, (that I think is also easier to manage
unsubscribing) maybe we don't need any more `openToNewsletter` flag in the User model...

estyxx:
Just to recap this: 
We would like to have a component in the Admin to filter and do some query to find a specific target of users to send communications (like reminders and so on...). We would like to use Amazon SES (Simple Email Service) for sending mass emails. But this "tool" doesn't keep a list of subscribers user like Mailchimp does. 

We just have the `openToNewsletter` property in the user model, so we can use that field to filter users willing to receive the newsletter. But we, of course, would like to send the newsletter to unregistered users; in the home page, we have the form where any person can put his email end be able to receive Newsletter without beeing register in the site. 

I think we can create an app where to put a little model for store the list of email addresses (we should also include registered users' mail: in the update Mutation of the user, if  `openToNewsletter` is true we should save the email also in this model/table; if it's false should remove!.  
And we can customize the admin of this app to send direct emails, if we want to send a Newsletter we will select only the users in the Newsletter-Table, otherwise make a query for a specific target...

estyxx:
#1300 I start a draft of this, but now I was working on Submission tag, maybe this weekend I will come back to this... :)

PR #1300, opened 2019-11-20
Merge pull request #1300 from pythonitalia/feature/newsletter

estyxx:
### TODO 
- - [ ] Understand how to pass email to unsubscribe to the FE [maybe pinpoint can help us]
- - [ ] Check the unsubscriber page message, now there is a simple and cold  "You have been successfully unsubscribed from this newsletter. ", do we want to put something like "we are sorry..." or "Did you click on unsubscribe by mistake? Click here to subscribe again!"... 
- - [ ] Retrieve these list of recipients: 
-  - [ ] Speakers without tickets
   -  [ ] Users that did not fill their ticket
   -  [ ] all users with ticket 
   -  [ ] Speakers that did not fill their ticket
- - [ ] Send list of subscriber to amazon pinpoint


marcoacierno:
Do we want to integrate with Mailchimp?

patrick91:
@marcoacierno I think we should migrate away from it, as it is quite expensive. We now have a big limit with SES, so we could use that to send emails, but we need to build something on django I guess

estyxx:
I think that this PR could be already merged. I add a component to keep the list of subscribers (registered user and not) and add a little form in the homepage in order to allow people to subscribe. 
We will have to integrate AWS and build a little tool (in the admin panel maybe) to allow us to send emails to a selected list of users (newsletter subscribers, but also specific user) see my comment [#1090](https://github.com/pythonitalia/pycon/issues/1090).

estyxx:
The PR it's almost complete... but continue to throw this error (in checks and locally): 
Screenshot 2019-12-09 at 22 12 25

I don't understand, seems there is some problem in that query when gatsby tries to execute in the start script
```
  const result = await graphql(
    `
      query {
        backend {
          blogPosts {
            slugEn: slug(language: "en")
            slugIt: slug(language: "it")
          }
          pages {
            slugEn: slug(language: "en")
            slugIt: slug(language: "it")
          }
        }
      }
    `,
  );
```

I have at least one blogPost and a Page, so there is something...

estyxx:
I added the support to send email from the admin panel. Maybe the template for the newsletter needs to be updated (now is the same of the change-password email...). 
And there is _that error_ I mentioned above, I don't understand what is happening...

backend/api/newsletters/forms.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class UnsubscribeToNewsletterForm(FormWithContext):
+    email = forms.EmailField()
+
+    def save(self):
+        email = self.cleaned_data.get("email")
+        try:
+            deleted, _ = Subscription.objects.get(email=email).delete()
+            return deleted == 1
```
patrick91: we can just have this, `Subscription.objects.get(email=email).delete()` no need to check :)

PR #1307, opened 2019-11-23
Tickets

marcoacierno:


marcoacierno:
image

backend/api/pretix/types.py
```diff
@@ -0,0 +1,72 @@
...
+class TicketItem:
+    id: strawberry.ID
+    name: str
+    description: Optional[str]
+    active: bool
+    default_price: str
+    variations: List[ProductVariation]
+    # TODO: correct types
+    available_from: Optional[str]
+    available_until: Optional[str]
```
marcoacierno: I don't think we have the DateTime type implemented available?

backend/api/tests/schema/test_create_order.py
```diff
@@ -0,0 +1,27 @@
...
+                "email": "patrick.arminio@gmail.com",
+                "locale": "en",
+            },
+        },
+    )
+
+    assert not response.get("errors")
+    assert response["data"]["createOrder"]["paymentUrl"] == "https://example.com"
+
+    create_order_mock.assert_called_once()
```
marcoacierno: nitpick: maybe we want to use assert_called_with making sure pretix is called with the correct tickets?

backend/importer/management/commands/import_old_db.py
```diff
@@ -615,6 +615,7 @@ def import_tickets(self, overwrite):
                         orig_ticket["ticket_user"] or None, order_user
                     )
 
+                    # TODO: import to pretix?
```
marcoacierno: def. something we can leave for later honestly

backend/pretix/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+
+def get_items(conference):
+    response = pretix(conference, "items")
+    response.raise_for_status()
+
+    data = response.json()
+    return {result["id"]: result for result in data["results"]}
+
+
+@strawberry.input
```
marcoacierno: do we want to move the strawberry's types in a different files in `api/`?

backend/pretix/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+        "positions": positions,
+    }
+
+    # it needs the / at the end...
+    response = pretix(conference, "orders/", method="post", json=payload)
+
+    if response.status_code == 400:
+        logger.warning("Unable to create order on pretix %s", response.content)
+
+    if not response.ok:
```
marcoacierno: I don't think you need the check? raise_for_status should throw only when the error is 400

backend/pretix/tests/conftest.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+from users.tests.fake_pretix import FAKE_PRETIX_ITEMS
+
+
+@pytest.fixture
```
marcoacierno: aah much better 👏

backend/pycon/settings/base.py
```diff
@@ -172,3 +169,16 @@
...
 SERIALIZATION_MODULES = {"json": "i18n.serializers"}
 USE_X_FORWARDED_HOST = False
+
+PRETIX_API = env("PRETIX_API", default="")
+PRETIX_API_TOKEN = None
+
+if PRETIX_API:
+    PRETIX_API_TOKEN = env("PRETIX_API_TOKEN")
+
+LOGGING = {
```
marcoacierno: we can configure this for prod to use cloudwatch in the future

PR #1388, opened 2019-12-07
Update Submission

marcoacierno:
This PR also expands Strawberry's "FormMutation" to support updating existing instances.


patrick91:
Added a couple of minor changes :)

PR #1404, opened 2019-12-14
Fix language picker not working on dynamic pages

marcoacierno:
Right now when you are on http://localhost:4000/it/submission/20 and try to switch to English (for example), you get redirected to `http://localhost:4000/it/submission` instead of `http://localhost:4000/en/submission/20` because the language picker uses the URL generated by gatsby at build time, which doesn't/can't know about the submission you are in.

This PR keeps the current behavior at build time, but when running in the client it uses the current browser URL to generate better language URLs.

### DIFF
--- a/frontend/src/helpers/wrap-page-element.tsx
+++ b/frontend/src/helpers/wrap-page-element.tsx
@@ -73,6 +73,29 @@ const isSocial = (props: Props["props"]) => {
   );
 };
 
+const getAlternateLinks = (props: Props["props"]) => {
+  if (typeof window !== "undefined") {
+    /*
+      when we are not SSO,
+      use the location's pathname to generate
+      more specific URLs:
+
+      Example /en/submission/{ID} cannot be generated server side,
+      so we replace it with the current pathname.
+    */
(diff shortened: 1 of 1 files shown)


frontend/src/helpers/wrap-page-element.tsx
```diff
@@ -73,6 +73,29 @@ const isSocial = (props: Props["props"]) => {
...
+      when we are not SSO,
+      use the location's pathname to generate
+      more specific URLs:
+
+      Example /en/submission/{ID} cannot be generated server side,
+      so we replace it with the current pathname.
+    */
+    const pathname = window.location.pathname
+      .replace("en", "")
+      .replace("it", "");
```
patrick91: wouldn't be better to replace "/en" and "/it"? :)

marcoacierno: yes

marcoacierno: I went for /en/ and /it/ so that to remove the entire initial part

PR #1407, opened 2019-12-15
Add new `speaker level` and `previous talk` to the submission

marcoacierno:


marcoacierno:
@patrick91 thanks 🙌 for spotting a bug where if the user doesn't select a choice in the speaker level (both when editing an old submission or creating a new one) they were unable to submit the submission

backend/api/tests/schema/test_edit_submission.py
```diff
@@ -133,6 +143,8 @@ def test_update_submission(
         new_tag=new_tag,
         new_duration=new_duration,
         new_type=new_type,
+        new_speaker_level=Submission.SPEAKER_LEVELS.experienced,
+        new_previous_talk_video="https://www.youtube.com/watch?v=dQw4w9WgXcQ",
```
patrick91: 😆

PR #1422, opened 2019-12-20
Log GraphQL errors

estyxx:
Add CustomGraphQLView to log errors when the request returns 400

### DIFF
--- a/backend/api/views.py
+++ b/backend/api/views.py
@@ -6,7 +6,7 @@ from django.http.response import HttpResponseBadRequest
 from django.shortcuts import render
 from graphql import graphql_sync
 from graphql.error import format_error as format_graphql_error
-from strawberry.contrib.django.views import GraphQLView
+from strawberry.django.views import GraphQLView
 
 
 class CustomGraphQLView(GraphQLView):
--- a/backend/api/views.py
+++ b/backend/api/views.py
@@ -52,7 +52,7 @@ class CustomGraphQLView(GraphQLView):
             ]
             self._capture_sentry_exceptions(result.errors)
 
-        return JsonResponse(response_data, status=400 if result.errors else 200)
+        return JsonResponse(response_data, status=200)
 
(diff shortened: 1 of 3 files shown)


backend/api/views.py
```diff
@@ -0,0 +1,19 @@
...
+        resp = super(CustomGraphQLView, self).dispatch(request, *args, **kwargs)
+        errors = json.loads(resp.getvalue()).get("errors", [])
+        if errors:
+            self._capture_sentry_exceptions(errors)
+        return resp
+
+    def _capture_sentry_exceptions(self, errors):
+        for error in errors:
+            logging.error(error, exc_info=True)
+            sentry_sdk.capture_exception(error["message"])
```
estyxx: Here I didn't put any try-except, as was doing graphene-sentry, I think strawberry will always give us "message" inside any error... @patrick91

backend/pycon/urls.py
```diff
@@ -1,14 +1,19 @@
...
 from django.conf import settings
 from django.conf.urls.static import static
 from django.contrib import admin
 from django.urls import include, path
 from django.views.decorators.csrf import csrf_exempt
-from strawberry.contrib.django.views import GraphQLView
+
+from api.schema import schema
+
+# from strawberry.contrib.django.views import GraphQLView
```
patrick91:
```suggestion
```

backend/api/views.py
```diff
@@ -0,0 +1,61 @@
...
+            self._capture_sentry_exceptions(result.errors)
+
+        return JsonResponse(response_data, status=200)
+
+    def _capture_sentry_exceptions(self, errors):
+        for error in errors:
+            try:
+                logging.error(error.original_error)
+            except AttributeError:
+                logging.error(error, exc_info=True)
```
patrick91: you can do `if hasattr('original_error')` instead of the try/except? :)

PR #1437, opened 2019-12-26
Ask for ticket type before selecting tickets

patrick91:


frontend/src/components/tickets-form/index.tsx
```diff
@@ -9,36 +9,58 @@ import { SelectedProductsWithVariationsList } from "./selected-products-with-var
...
-    />
-  </React.Fragment>
-);
+}) => {
+  const ticketsToShow = tickets.filter(ticket => {
+    if (ticket.variations!.length > 0) {
+      return true;
+    }
+
+    if (isBusiness && ticket.name.toLowerCase().includes("business")) {
```
marcoacierno: I fear we do not have a better way to filter business and not business tickets :( @patrick91?

frontend/src/components/tickets-form/index.tsx
```diff
@@ -9,36 +9,58 @@ import { SelectedProductsWithVariationsList } from "./selected-products-with-var
...
+
+  return (
+    <React.Fragment>
+      {ticketsToShow.map(ticket => (
+        <Box key={ticket.id}>
+          <ProductRow
+            quantity={
+              selectedProducts[ticket.id]
+                ? selectedProducts[ticket.id].length
+                : 0
```
marcoacierno: this is something I wrote do you think we can do `selectedProducts?.length ?? 0`?

marcoacierno:
Looks fine to me! 

There is a typescript error:

```
src/components/tickets-page/index.tsx(43,7): error TS2322: Type 'string' is not assignable to type 'boolean'.
```

frontend/src/components/tickets-form/index.tsx
patrick91: we could use categories (and have a flag on graphql, so we do this check on the backend)

PR #1440, opened 2019-12-30
Voting

marcoacierno:


backend/api/voting/forms.py
```diff
@@ -33,3 +36,43 @@ def save(self, commit=True):
     class Meta:
         model = Vote
         fields = ("submission", "value")
+
+
+class SendVotesForm(ContextAwareForm):
```
estyxx: Why we need two Votes form? If we want to save the votes altogether, ok, but can we use only one form, and pass only one item if we save a single vote?

marcoacierno: Yes, we do not need the two forms :) The plan is to batch update the votes instead of sending N requests.

Probably it's worth adding a reminder to delete the old mutation once the PR is complete

marcoacierno: We changed how the saving works and now we use the single `sendVote` only 🎉

backend/voting/models.py
```diff
@@ -7,10 +7,11 @@
...
     VALUES = Choices(
-        (0, "not_interested", _("Not Interested")),
-        (1, "maybe", _("Maybe")),
-        (2, "want_to_see", _("Want to See")),
-        (3, "must_see", _("Must See")),
+        (1, "not_interested", _("Not Interested")),
+        (2, "maybe", _("Maybe")),
+        (3, "want_to_see", _("Want to See")),
+        (4, "must_see", _("Must See")),
+        (5, "love_it", _("Love it")),
```
marcoacierno: let's find a better name for this :P @patrick91 or maybe just use Integers instead of ChoiceField? we do not use the text anywhere...

backend/voting/tests/test_forms.py
```diff
@@ -0,0 +1,31 @@
+from api.helpers.ids import encode_hashid
+from api.voting.forms import HashidModelChoiceField
+from django.forms import ValidationError
+from pytest import mark, raises
+from submissions.models import Submission
+
+
+@mark.django_db
+def test_hashid_model_choice_field(submission):
+    field = HashidModelChoiceField(queryset=Submission.objects.all())
```
marcoacierno: I think we should create a custom manager named `HashidManager` that has support for hashids, in case we want to use it elsewhere too. Probably something for another PR

PR #1442, opened 2019-12-31
Hotels

marcoacierno:


backend/api/hotels/types.py
```diff
@@ -0,0 +1,12 @@
...
+from ..helpers.i18n import make_localized_resolver
+
+
+@strawberry.type
+class HotelRoom:
+    id: str
+    name: str = strawberry.field(resolver=make_localized_resolver("name"))
+    description: str = strawberry.field(resolver=make_localized_resolver("description"))
+    price: int
+    is_sold_out: bool
```
patrick91: I'd do a quantity left, so we prevent to book more rooms than are available, but this can be dealt with later (we are not going to have a lot of people booking for more than one room)

backend/api/orders/mutations.py
```diff
@@ -20,15 +24,28 @@ class Error:
 
 @strawberry.type
 class OrdersMutations:
-    @strawberry.mutation
+    @strawberry.mutation(permission_classes=[IsAuthenticated])
```
patrick91: 🙇

backend/api/orders/mutations.py
```diff
@@ -38,3 +55,44 @@ def create_order(
...
         return CreateOrderResult(payment_url=payment_url)
+
+
+def validate_hotel_rooms(hotel_rooms: typing.List[CreateOrderHotelRoom], *, conference):
+    for order_room in hotel_rooms:
+        try:
+            room = HotelRoom.objects.get(id=order_room.room_id)
+        except HotelRoom.DoesNotExist:
+            return Error(
+                message=_("Room %(id)s not found") & {"id": order_room.room_id}
```
patrick91: why &? shouldn't it be %?

backend/api/orders/mutations.py
```diff
@@ -38,3 +55,44 @@ def create_order(
...
+        if room.is_sold_out:
+            return Error(
+                message=_("Room %(id)s is sold out") & {"id": order_room.room_id}
+            )
+
+        conference_start = conference.start.date()
+        conference_end = conference.end.date()
+
+        if order_room.checkin < conference_start or order_room.checkin > conference_end:
+            return Error(message=_("Invaild checkin date"))
```
patrick91:
```suggestion
            return Error(message=_("Invalid check-in date"))
```

backend/api/orders/mutations.py
```diff
@@ -38,3 +55,44 @@ def create_order(
...
+
+        if (
+            order_room.checkout < conference_start
+            or order_room.checkout > conference_end
+            or order_room.checkin > order_room.checkout
+        ):
+            return Error(message=_("Invaild checkout date"))
```
patrick91:
```suggestion
            return Error(message=_("Invalid check-out date"))
```

backend/api/orders/mutations.py
```diff
@@ -38,3 +55,44 @@ def create_order(
...
+
+
+def create_hotel_reservation(
```
patrick91: I'd rename this to create_hotel_reservations (plural) since it creates multiple reservations :)

marcoacierno: done!

backend/api/orders/mutations.py
marcoacierno: done!

backend/api/orders/mutations.py
marcoacierno: done!

backend/api/orders/mutations.py
marcoacierno: lol ops

backend/api/hotels/types.py
marcoacierno: hmm I see, it makes sense to block if the user selects like 30 rooms and we have only 25. I will add a check that the number of rooms you are buying is less than the available quantity

backend/api/orders/mutations.py
marcoacierno: I am going to add a test case for this

backend/api/hotels/types.py
marcoacierno: I added a check so that the user cannot buy  200 rooms in one go when the capacity is 50

backend/api/orders/mutations.py
marcoacierno: Added

backend/api/hotels/types.py
marcoacierno: done frontend check

PR #1464, opened 2020-01-15
generate submission ranking from django-admin

estyxx:
[ ] Add field to know if the vote was normalized during the ranking phase True / False 
[ ] If the user didn't vote a submission assign "maybe" as default, in order to normalize and not create a disadvantage
[ ] Save the result generated-ranking [Run, SubmissionRank]
[ ] Create a page to visualize all this nice data :) with 
 - Link to the submission


backend/voting/models/ranking.py
```diff
@@ -0,0 +1,171 @@
...
+                "score": score,
+            }
+            ranking.append(rank)
+        return sorted(ranking, key=lambda k: k["score"], reverse=True)
+
+    @staticmethod
+    def get_users_weights(conference):
+        from users.models import User
+
+        users = User.objects.all()
```
leriomaggio: To reduce the hit to the Database and retain only interesting users I would rather use the `submission` query set and get all the unique users from that.
I don't know if this **actually** works - should be tested - but something like:

```
user_ids = submissions.distinct('user').values_list('user', flat=True)
voters = User.objects.filter(id__in=user_ids)
```

PR #1482, opened 2020-01-26
Fix import old db

estyxx:


marcoacierno:
@estyxx do you remmeber why we need those changes?

estyxx:
> @estyxx do you remmeber why we need those changes?

Yes, we made changes to some models and the import wasn't updated. Also, I added old tags and votes. We imported only the users the other entities were never imported.

PR #1498, opened 2020-02-01
Implement new design for the profile page

marcoacierno:
New profile page according to the design

The avatar and "My tickets" section is missing and we can implement it in another PR

# How to test

Open Vercel preview, login with your production account and see the new profile. Try editing and all

image

this PR already has the correct structure to implement the avatar section later

orders table:

(desktop)

image

(mobile)

image

I also changed how My submission looks

This is in the profile page:

image

This is in the CFP page:

image


frontend/src/app/profile/logout.tsx
```diff
@@ -44,9 +44,10 @@ export const Logout: React.SFC<{ lang: string }> = ({ lang }) => {
           px: 3,
         }}
       >
-        <Heading mb={2} as="h2">
+        <Heading mb={2} as="h1">
```
patrick91: did you do this to change the font size?

marcoacierno: not really, more a side effect while I was experimenting with refactoring the code a bit, I want to make everything an h2 and the main one h1.

SEO-wise and dom-wise(? :P) it's best to have the `h1` for `My Profile` as it's the title of the page and `h2` for the others

patrick91: sounds good, just wondering if the font size will come from theme ui or from the UA stylesheet :D

marcoacierno: in the theme we have the default style of `h1` etc so it will be from theme-ui

marcoacierno: image

backend/pretix/db.py
```diff
@@ -30,3 +34,132 @@ def user_has_admission_ticket(email: str, event_slug: int):
...
+            positions[position_id]["answers"].append(
+                {"question_id": answer[0], "answer": answer[1]}
+            )
+
+        # This approach tries to fetch everything, answers AND questions list
+        # this does not support fetching a question's options/choices
+        # (those are in another table)
+        # this way allows the user to get everything without asking external services
+        # or other things
+
```
marcoacierno: @patrick91 I am a bit conflicted about the best way to proceed here.

What we have to do:

Fetch the user's answers for each ticket (=position) assigned to them.

How we can do it:

1. Fetch from the Pretix database all the answers of that position, without any information about the question (title, type of question, required etc.) This means that if the frontend wants to render the UI they have to fetch the questions' information. Example:

```graphql
{
  conference(code: "pycon-demo") {
    tickets(language: "en") {
      id
      questions {
        id
        required
        name
        options {
          name
          id
        }
      }
    }
  }
  
  me {
    tickets(conference: "pycon-demo", language: "en") {
      name
      positionId
      
      answers {
        questionId
        answer
      }
    }
  }
}```

This makes the backend send an HTTP request to the pretix API to get the items information and two DB requests to get the answers and user's tickets.

This is the easiest to implement (see the currently active, not commented code)

2. Another way is to fetch question+answer together (commented code), the advantage of this solution is that no API requests are being made but the SQL query is a bit of a mess (we also need to add another query to get all the possible options of a question)


I am pretty sure there are other solutions for this as well, but it might be too late for me right now 💤

Let me know what do you think @patrick91 if you can think of a better solution or anything, thanks!

marcoacierno:
It seems pretix does not have an API to update a order position values, so we might need to resort to running a query in the database to change the answers via frontend

estyxx (approved):

PR #1520, opened 2020-02-14
Save users' interests in ScheduleItem

estyxx:


backend/schedule/models.py
```diff
@@ -131,6 +131,13 @@ class ScheduleItem(TimeStampedModel):
         on_delete=models.PROTECT,
     )
 
+    subscribed_users = models.ManyToManyField(
```
marcoacierno, line 134: quick look and a nitpick: I am not sure we should save the preferences on the schedule item, I would keep the user part and the schedule item separated

PR #1547, opened 2020-03-16
Implement workshop booking

marcoacierno:


backend/api/schedule/mutations.py
```diff
@@ -42,8 +45,70 @@ class UpdateOrCreateSlotItemResult:
...
+        self, info, id: strawberry.ID, conference: str
+    ) -> typing.Union[ScheduleItemBookingStatus, ReleaseScheduleItemBookingError]:
+        user = info.context["request"].user
+
+        schedule_item_booking = ScheduleItemBooking.objects.filter(
+            user=user, schedule_item_id=id, schedule_item__conference__code=conference
+        ).first()
+
+        if not schedule_item_booking:
+            return ReleaseScheduleItemBookingError(message=_("No booking found"))
```
patrick91: this could a BookingNotFound error ;)

patrick91: same for the other ones, I imagine the FE using the typename information to show an appropriate message 😊

backend/api/schedule/types.py
```diff
@@ -35,6 +37,25 @@ class ScheduleItem:
...
+    @strawberry.field
+    def is_booked(self, info) -> bool:
+        user = info.context["request"].user
+
+        if not user.is_authenticated:
+            return False
+
+        return ScheduleItemBooking.objects.filter(
+            user=user, schedule_item=self
+        ).exists()
```
patrick91: we need a way to optimise this at some point 😊

backend/api/schedule/types.py
```diff
@@ -35,6 +37,25 @@ class ScheduleItem:
...
+
+    @strawberry.field
+    def can_book(self, info) -> bool:
+        return self.allows_booking and self.capacity_left > 0
```
patrick91: we could reuse the resolver 😊

```suggestion
    def can_book(self, info) -> bool:
        return self.allows_booking and self.has_free_spot(info)
```

patrick91: we could also move the resolver to be a property on the model, but that's minor

backend/schedule/models.py
```diff
@@ -131,6 +133,21 @@ class ScheduleItem(TimeStampedModel):
         on_delete=models.PROTECT,
     )
 
+    maximum_capacity = models.PositiveIntegerField(
+        _("maximum capacity"), blank=True, null=True
+    )
+
+    @property
+    def allows_booking(self):
+        return self.type == ScheduleItem.TYPES.training
```
patrick91: we need to take into account also SUBMISSION types where the submission is a training :)

backend/schedule/models.py
```diff
@@ -131,6 +133,21 @@ class ScheduleItem(TimeStampedModel):
...
+
+    @property
```
patrick91: let's make this a cached property :)

marcoacierno: done

backend/api/schedule/types.py
marcoacierno: moved it to the model

PR #1724, opened 2020-08-07
Update strawberry

marcoacierno:
Update strawberry to latest version


backend/api/blog/types.py
```diff
@@ -2,7 +2,7 @@
 
 import strawberry
 from api.users.types import User
-from strawberry.types.datetime import DateTime
+from strawberry.schema.types.base_scalars import DateTime
```
patrick91, line 5: I'd use datetime from python directly :)

backend/api/events/types.py
```diff
@@ -1,20 +1,18 @@
...
 
-if TYPE_CHECKING:  # pragma: no cover
-    from api.conferences.types import Conference
-
 
 @strawberry.type
 class Event:
     id: strawberry.ID
-    conference: "Conference"
+    conference: LazyType["Conference", "api.conferences.types"]
```
patrick91, line 15: can you try with:

```suggestion
    conference: "api.conferences.types.Conference"
```

just curious to see if this works, maybe LazyType is actually not needed :)

backend/strawberry_forms/converter.py
```diff
@@ -33,15 +33,15 @@ def type_or_optional_wrapped(type_, required):
 def convert_form_field_to_string(field):
     return (
         type_or_optional_wrapped(str, field.required),
-        strawberry.field(description=field.help_text, is_input=True),
+        strawberry.input(description=field.help_text),
```
patrick91, line 36: nnot sure this is right, why did you change it?

backend/strawberry_forms/mutations.py
```diff
@@ -101,9 +102,11 @@ def mutate(root, info, input: input_type) -> output:
             def mutate(root, info) -> output:
                 return _mutate(root, info, {})
 
-        mutate.__name__ = f"{name}Output"
-        mutate.name = name
-        mutate = strawberry.mutation(mutate, description=cls.__doc__)
+        mutate = strawberry.mutation(
+            mutate,
+            name=f'{name[0].lower()}{name[1:]}',
+            description=cls.__doc__
```
patrick91, line 108: let's not use descriptioon from the doc string ;)

backend/api/events/types.py
marcoacierno: it works but I need to `import api` at the top, I am not sure I like it 🤔

backend/strawberry_forms/converter.py
```diff
@@ -33,15 +35,15 @@ def type_or_optional_wrapped(type_, required):
 def convert_form_field_to_string(field):
     return (
         type_or_optional_wrapped(str, field.required),
-        strawberry.field(description=field.help_text, is_input=True),
+        dataclasses.field(default=field.initial),
```
marcoacierno, line 38: we lose the `help_text` support here because we need to set the default value coming from the form, but as I plan on removing the forms integration anyway I think that's ok

backend/strawberry_forms/mutations.py
```diff
@@ -101,9 +104,10 @@ def mutate(root, info, input: input_type) -> output:
             def mutate(root, info) -> output:
                 return _mutate(root, info, {})
 
-        mutate.__name__ = f"{name}Output"
-        mutate.name = name
-        mutate = strawberry.mutation(mutate, description=cls.__doc__)
+        # Hack because something changed and now the name contains `Mutation`
+        # as the classname, which makes sense but why it didn't happen before? lol
+        name = name.replace("Mutation", "")
```
marcoacierno, line 109: so the name of the class is `ResetPasswordMutation`, assigned to `reset_password = ResetPasswordMutation`

Before the update of strawberry the mutation name used to be `resetPassword` but now it is `resetPasswordMutation`

This is a quick hack as I plan on removing the django forms integration anyway :)

backend/users/tests/test_reset_password_flow.py
```diff
@@ -138,7 +138,7 @@ def test_cannot_reset_password_of_an_unkown_user(user_factory, graphql_client):
     user = user_factory(password="old")
     token = default_token_generator.make_token(user)
 
-    response = _reset_password(graphql_client, token, 5, "another")
+    response = _reset_password(graphql_client, token, 0, "another")
```
marcoacierno, line 141: this test seems wrong, to use `5` like 🤷 , these tests need a refactor as well

backend/strawberry_forms/tests/test_mutations.py
```diff
@@ -1,73 +0,0 @@
-import strawberry
```
marcoacierno, line 1: I didn't manage to get the mutation to work, if someone has any idea happy to hear it, otherwise, I would just delete this code and move on removing the rest of strawberry_forms

backend/api/events/types.py
patrick91: you can hide it under `if TYPE_CHECKING`, anyway both solutions are ok for now :)

patrick91 (approved):

PR #1908, opened 2020-12-03
Re-implement reset password as DDD

marcoacierno:
Still a WIP should give an idea on how we might implement stuff using DDD :)

### DIFF
--- a/backend/users/domain/entities/user.py
+++ b/backend/users/domain/entities/user.py
@@ -1,0 +1,42 @@
+from datetime import date, datetime
+from enum import Enum
+from dataclasses import dataclass
+from typing import NewType, Optional
+
+
+UserID = NewType("UserID", int)
+
+
+class Gender(Enum):
+    MALE = "male"
+    FEMALE = "female"
+    OTHER = "other"
+    NOT_SAY = "not_say"
+
+
+@dataclass
+class User:
+    id: UserID
+    username: str
+    email: str
+    full_name: str
+    name: str
+    gender: Gender
+    date_of_birth: date
+    country: str
+    open_to_recruiting: bool
+    open_to_newsletter: bool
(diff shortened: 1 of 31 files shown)


backend/users/domain/entities/user.py
```diff
@@ -0,0 +1,42 @@
...
+    # Hashed password
+    password: Optional[str] = None
+    new_password: Optional[str] = None
+
+    def set_password(self, password: str):
+        self.new_password = password
+
+    @property
+    def pk(self) -> UserID:
+        return self.id
```
patrick91, line 42: do you need this? seems like you're leaking implementation details here :)

backend/users/domain/repository.py
```diff
@@ -0,0 +1,50 @@
...
+    def get_by_id(self, user_id: UserID) -> Optional[User]:
+        user = models.User.objects.filter(id=user_id).first()
+
+        if not user:
+            return None
+
+        return convert_user(user)
+
+    def validate_reset_password_token(self, user: User, token: str) -> bool:
+        return default_token_generator.check_token(user, token)
```
patrick91, line 19: this feels likes it doesn't really belong here

marcoacierno: I made the `User` entity similar enough to the django model so that `default_token_generator.check_token` can work with the entity and not the model, so I can move it outside the repository

otherwise we would need the django object to call `check_token`

backend/users/domain/entities/user.py
marcoacierno: https://github.com/pythonitalia/pycon/pull/1908#discussion_r535731422

this was done so we can call `check_token` passing the entity instead of the django model, I can't think of a better way to do it 🤔

PR #1910, opened 2020-12-04
Migrate backend to AWS Lambda

marcoacierno:


.github/workflows/deploy.yml
```diff
@@ -80,3 +97,15 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [terraform]
+    steps:
+      - uses: actions/checkout@master
+      - name: Migrate DB
+        run: 'aws lambda invoke --function-name production-pycon-backend --cli-binary-format raw-in-base64-out --payload ''{ "_serverless-wsgi": { "command": "manage", "data": "migrate" } }'''
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
patrick91, line 111: nice!

backend/Dockerfile.lambda
```diff
@@ -0,0 +1,69 @@
...
+    /usr/lib/liblber-2.4.so.2 \
+    /usr/lib/libsasl2.so.3 \
+    /usr/lib/libxslt.so.1 \
+    /usr/lib/libexslt.so.0 \
+    /usr/lib/libxml2.so.2 \
+    /usr/lib/libgcrypt.so.20 \
+    /usr/lib/libgpg-error.so.0 \
+    /usr/lib/libstdc++.so.6 \
+    /usr/lib/libgcc_s.so.1 \
+    /usr/lib/
```
patrick91, line 56: ahaha, well done 👏  

the last line refers to the destination, right?

infrastructure/backend_ecr.tf
```diff
@@ -0,0 +1,41 @@
...
+{
+    "rules": [
+        {
+            "rulePriority": 1,
+            "description": "Expire images older than 1 days",
+            "selection": {
+                "tagStatus": "untagged",
+                "countType": "sinceImagePushed",
+                "countUnit": "days",
+                "countNumber": 1
```
patrick91, line 27: is this something we want?

patrick91 (approved):
Nice stuff! Added a minor comment :)

infrastructure/backend_ecr.tf
marcoacierno: Hmm I went for storing only the latest image, and delete anything older as we do a deploy every time we push to master so having all those docker images that we do not use seems a bit wasteful

We also need to improve our deployment to build/release the BE only when we change something there

patrick91: so it doesn't delete the last one?

marcoacierno: yup only untagged images, so when we release a new build the new image gets "latest" and the previous one untagged

backend/Dockerfile.lambda
marcoacierno: yup

PR #1930, opened 2020-12-18
Fix sponsors image in homepage

marcoacierno:
Update our Vercel website to not use imgix (we disable our subscription) replacing it with the internal next/image process


frontend/next.config.js
```diff
@@ -12,7 +12,9 @@ module.exports = withSourceMaps({
     conferenceCode: process.env.CONFERENCE_CODE || "pycon-demo",
     SENTRY_DSN: process.env.SENTRY_DSN || null,
   },
-
+  images: {
+    domains: ['production-pycon-backend-media.s3.amazonaws.com'],
+  },
```
patrick91, line 17: shall we add pycon.it as well?

patrick91 (approved):

frontend/next.config.js
marcoacierno: for assets coming from nextjs we shouldn't need to add the domain as the URL will be /XX

what use case are you thinking where we will have an URL like `pycon.it/media/image.png`?

Issue #1956, opened 2020-12-23
Update to Python 3.9


marcoacierno:
We are on Python 3.7 right now, we use Lambda with a Container so we can use any Python version :) so it would be great to be on the latest version

PR #1988, opened 2020-12-27
Update to python 3.9 (1956)

estyxx:
#1956

### DIFF
--- a/.github/workflows/backend-checks.yml
+++ b/.github/workflows/backend-checks.yml
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@master
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.9.0"
+          python-version: "3.9.1"
       - run: pip install poetry
 
       - name: Cache Poetry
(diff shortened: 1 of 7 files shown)


.github/workflows/backend-checks.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@master
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.7.9"
+          python-version: "3.9.0"
```
marcoacierno, line 17: can you update to 3.9.1?

marcoacierno (commented):
Overall looks good

Can you also update the python version in pre-commit (here: https://github.com/pythonitalia/pycon/blob/master/.pre-commit-config.yaml#L20)?

.github/workflows/backend-checks.yml
estyxx: done

backend/Dockerfile.lambda
```diff
@@ -1,6 +1,6 @@
 ARG FUNCTION_DIR="/home/app/"
 
-FROM python:3.7-alpine as build-stage
+FROM python:3.9-alpine as build-stage
```
marcoacierno, line 3: we have another `FROM python:3.7-alpine` below

also need to change

```
COPY --from=build-stage /usr/local/lib/python3.7/site-packages /usr/local/lib/python3.7/site-packages
```

marcoacierno (approved):
Thanks!

PR #2012, opened 2021-01-24
Users microservice POC

marcoacierno:


users/bin/makemigrations
```diff
@@ -0,0 +1 @@
+poetry run alembic revision --autogenerate "$@"
```
patrick91, line 1: I'd make a Makefile or use pyinvoke :) Not a big fan of these scripts 😊

patrick91 (approved):

users/main.py
```diff
@@ -0,0 +1,22 @@
+import databases
```
estyxx, line 1: Isn't better "app.py" for this? [but I don't know the conventions...]

users/bin/makemigrations
estyxx: agree :)

estyxx (approved):

users/bin/makemigrations
marcoacierno: I don't think we can use pyinvoke

I will have a look at a Makefile

users/main.py
marcoacierno: the module inside it already exposes `app` so it might be a bit repetitive

users/bin/makemigrations
marcoacierno: Changed to use `pyinvoke`, and to make it easier to write I did `make run`, so you can do

`make run migrate`
`make run makemigrations "Name migration"`
`make run server`

users/main.py
```diff
@@ -0,0 +1,35 @@
...
+@app.middleware("http")
+async def async_session_middleware(request, call_next):
+    async with get_session(request.app.state.engine) as session:
+        request.state.session = session
+
+        try:
+            return await call_next(request)
+        finally:
+            # TODO is needed?
+            request.state.session = None
```
patrick91: did you figure this out?

users/main.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+@app.on_event("startup")
+async def startup():
+    app.state.engine = get_engine()
```
patrick91, line 30: do we need a shutdown function?

users/users/api/tests/mutation/test_login_mutation.py
```diff
@@ -0,0 +1,170 @@
...
+            __typename
+        }
+    }
+    """
+    response = await graphql_client.query(
+        query, variables={"input": {"email": "hah@pycon.it", "password": "ciao"}}
+    )
+
+    assert not response.errors
+    assert response.data["login"]["__typename"] == "UsernameAndPasswordCombinationWrong"
```
patrick91, line 22:
```suggestion
    assert response.data["login"]["__typename"] == "WrongUsernameOrPassword"
```

users/main.py
marcoacierno: I removed it because I assumed the engine would get disposed automatically, but  I think you might be right, maybe when live-reloading it doesn't

I will add it again now

users/main.py
marcoacierno: had a look at the docs of starlette and seems that they should be cleaned up automatically

PR #2111, opened 2021-03-15
Refresh token logic

marcoacierno:


gateway/actions/auth-action.ts
```diff
@@ -1,23 +1,56 @@
...
+    // clear previous set cookies (clear identity cookies)
+    context.setCookies.splice(0, context.setCookies.length);
 
     context.setCookies.push({
       name: "identity",
       value: identityToken,
       options: {
         httpOnly: true,
-        maxAge: 60 * 15,
+        maxAge: 60 * 60 * 24 * 7 * 4 * 3,
```
patrick91, line 31: can we use a constant here to make it more readable? :)

gateway/handler.ts
```diff
@@ -1,29 +1,67 @@
...
-    const cookieHeader = event.headers["cookie"];
-    let identity = null;
+    return createContext(event.headers["cookie"]);
+  },
+});
 
-    if (cookieHeader) {
-      const cookies = cookie.parse(event.headers["cookie"]);
-      identity = cookies["identity"];
+const handleManyCookies = (headers: any = {}) => {
```
patrick91, line 16: what's this for? can we add some comments?😊

gateway/package.json
```diff
@@ -18,11 +17,17 @@
   },
   "scripts": {
     "dev": "NODE_ENV=development IS_DEV=true VARIANT=default nodemon",
-    "dev:admin": "NODE_ENV=development IS_DEV=true VARIANT=admin nodemon"
+    "dev:admin": "NODE_ENV=development IS_DEV=true VARIANT=admin nodemon",
```
patrick91, line 20: why do we need both node_env and is_dev?

gateway/pastaporto/__tests__/identity.test.ts
```diff
@@ -0,0 +1,291 @@
...
+    const expectedToken =
+      "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpYXQiOjE2MTYyNTA5MzEsImV4cCI6MTYyMzUwODUzMSwiYXVkIjoicmVmcmVzaCIsImlzcyI6ImdhdGV3YXkiLCJzdWIiOiIxMCJ9.8nF91uPV7ngknINxD0X9U58t-Z7JPwogQw9ngx6cNI4";
+
+    expect(createRefreshToken("10")).toBe(expectedToken);
+  });
+
+  test("Cannot create refresh token with empty subject", () => {
+    expect(() => createRefreshToken("")).toThrow("Empty subject not allowed");
+  });
+});
```
patrick91, line 261: awesome work, wouldn't it be best to use an utility function to generate these tokens? they are hard to scan/debug 😊

patrick91 (approved):

gateway/pastaporto/__tests__/identity.test.ts
marcoacierno: sure, will change. I guess we can assume that the utility that creates the token is always right, one of the reason I used pre-created tokens was so we would be safe from bugs in the jwt library itself

gateway/package.json
marcoacierno: 🤔 I assume to allow to start with production bundle without https

marcoacierno: I will change to use NODE_ENV for now then we will see

gateway/handler.ts
marcoacierno: done

gateway/actions/auth-action.ts
marcoacierno: maybe now it's more clear?

gateway/pastaporto/__tests__/identity.test.ts
marcoacierno: done

PR #2126, opened 2021-03-22
Reset password flow

marcoacierno:
image


users-backend/users/api/mutations/request_reset_password.py
```diff
@@ -0,0 +1,30 @@
...
+    user = await info.context.users_repository.get_by_email(email)
+
+    if not user:
+        logger.info("Trying to request reset password of not existent user")
+        return OperationSuccess(ok=True)
+
+    try:
+        await services.request_reset_password(user)
+    except UserIsNotActiveError:
+        pass
```
patrick91, line 28: why pass here? should we add some logs in any case?

users-backend/users/domain/entities.py
```diff
@@ -91,6 +112,13 @@ def display_name(self) -> str:
     Column("is_active", Boolean(), default=True, nullable=False),
     Column("is_staff", Boolean(), default=False, nullable=False),
     Column("is_superuser", Boolean(), default=False, nullable=False),
+    Column(
+        "request_reset_password_id",
```
patrick91, line 116: I'd make this a bit more generic as we can use also to invalidate existing login tokens (when someone logs out from everywhere, when they change password, when we ban them)

marcoacierno: sure

marcoacierno: consider that every time the user sends a request to our gateway we get information about the user, so it's easier for us invalidating tokens and so on

users-backend/users/api/mutations/request_reset_password.py
marcoacierno: we already log it in the service

users-backend/users/domain/entities.py
```diff
@@ -63,6 +65,24 @@ def set_password(self, raw_password: str):
     def has_usable_password(self) -> bool:
         return is_password_usable(self.hashed_password)
 
+    def get_reset_password_jwt_id(self) -> str:
```
marcoacierno, line 68: test this

users-backend/users/domain/entities.py
```diff
@@ -63,6 +65,24 @@ def set_password(self, raw_password: str):
...
+        return f"reset-password:{self.id}:{self.jwt_auth_id}"
+
+    def create_reset_password_token(self) -> str:
```
marcoacierno, line 71: test this

users-backend/users/domain/services/__init__.py
```diff
@@ -13,4 +15,9 @@
     "social_login",
     "SocialLoginInput",
     "SocialAccount",
+    # reset_password
+    "reset_password",
+    "ResetPasswordInput",
+    # request_reset_password
```
estyxx, line 21: this comments now can be removed....

PR #2194, opened 2021-05-02
Show login/register errors in ModalSigning

estyxx:


marcoacierno (approved):
my idea was to implement error messages when the password is too short etc this sort of cases but we can merge this as a start and implement that in another PR, what do you think @estyxx ?

PR #2195, opened 2021-05-02
Implement loading states when fetching users/association stripe link #2167

estyxx:


association-frontend/src/components/button/index.tsx
```diff
@@ -1,40 +1,59 @@
...
+  const [emojiIndex, setEmojiIndex] = useState(0);
+
+  useEffect(() => {
+    if (loading) {
+      const interval = setInterval(() => {
+        setEmojiIndex((emojiIndex + 1) % loadingEmojis.length);
+      }, 600);
+      return () => clearInterval(interval);
+    }
+  });
```
marcoacierno, line 34: you are missing the deps here meaning that this effect is called at every render, it seems that we want this only when `loading` changes?

so 

```suggestion
  }, [loading]);
```

marcoacierno:
happy to merge once the deps array is specified in the useEffect

marcoacierno (approved):
looks good

PR #2229, opened 2021-05-26
Setup Apollo studio on the gateway

marcoacierno:


.github/actions/start-services/action.yml
```diff
@@ -0,0 +1,84 @@
+name: 'Start all microservices'
```
marcoacierno, line 1: I think this will look better once we have a good docker-compose local setup

infrastructure/applications/gateway/main.tf
```diff
@@ -19,15 +14,15 @@ module "lambda" {
   docker_repository_name = "gateway"
   docker_tag             = terraform.workspace
   role_arn               = data.aws_iam_role.lambda.arn
+  memory_size            = 1024
```
marcoacierno, line 17: kind of hoping that giving it more memory would improve cold start, but it seems that cold start with docker images aren't great.

need more investigation, 1gb isn't a lot so it's ok to keep it I think. we could raise all services from 512mb to 1gb to be honest

gateway/handler.ts
```diff
@@ -50,24 +52,28 @@ const handleManyCookies = (headers: any = {}) => {
   return { headers };
 };
 
+let serverHandler: ReturnType<ApolloServer["createHandler"]> | null = null;
```
marcoacierno, line 55: moved it outside so that everything the graphqlHandler is called we do not have to create a new handler, should improve performance

gateway/gateway.ts
```diff
@@ -49,16 +49,24 @@ class ServiceRemoteGraphQLDataSource extends RemoteGraphQLDataSource {
...
+export const createGateway = () => {
+  const options: any = {};
 
-    return new ServiceRemoteGraphQLDataSource({
-      url,
-    });
-  },
-});
+  if (IS_DEV) {
+    options.serviceList = getServices();
```
marcoacierno, line 56: on dev for now we still have a local list of services

maybe in the future we could use apollo studio development schemas, but it seems overkill

backend/Dockerfile.lambda
```diff
@@ -54,6 +54,7 @@ COPY --from=build-stage /usr/lib/libpq.so.5 \
     /usr/lib/libgpg-error.so.0 \
     /usr/lib/libstdc++.so.6 \
     /usr/lib/libgcc_s.so.1 \
+    /usr/lib/libjpeg.so.8 \
```
marcoacierno, line 57: this fixed an error on production that you cannot upload images

frontend/src/pages/[lang]/profile/edit/index.tsx
```diff
@@ -82,10 +82,9 @@ export const EditProfilePage: React.FC = () => {
   const router = useRouter();
   const language = useCurrentLanguage();
   const [loggedIn] = useLoginState();
-  const [
-    formState,
-    { text, select, checkbox, raw },
-  ] = useFormState<MeUserFields>(
+  const [formState, { text, select, checkbox, raw }] = useFormState<
+    MeUserFields
+  >(
```
Comment (marcoacierno) [inline frontend/src/pages/[lang]/profile/edit/index.tsx:87]: prettier misconfiguration between pre-commit and local prettier

we need to look at it and fix it

gateway/Dockerfile
```diff
@@ -21,6 +21,9 @@ WORKDIR ${LAMBDA_TASK_ROOT}
 COPY --from=build /app/node_modules ./node_modules
 COPY --from=build /app/dist ./
 
+# Copy some static files
+COPY --from=build /app/internal-services/logout/schema.graphql ./internal-services/logout/schema.graphql
```
marcoacierno, line 25: we only copy the typescript build above (/app/dist), files that are not ts are ignored so we need to manually copy them

.github/workflows/deploy.yml
```diff
@@ -191,3 +193,102 @@ jobs:
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
+
+  update-apollo-studio-graphs:
```
marcoacierno, line 197: idea: to make this quicker we could start and create the *.graphql files as initial step of the deploy

and after terraform just publish them

something similar to what we do for the emails

PR #2264, opened 2021-07-01
Locale improvements

marcoacierno:


frontend/src/locale/context.tsx
```diff
@@ -22,8 +24,11 @@ export const LocaleProvider: React.FC<{ lang: string }> = ({
   children,
 }) => {
   React.useEffect(() => {
-    if (lang !== localStorage.getItem("locale")) {
-      localStorage.setItem("locale", lang);
+    const { pyconLocale } = cookies({
+      req: { headers: { cookie: document.cookie } },
+    });
+    if (lang !== pyconLocale) {
+      document.cookie = `pyconLocale=${lang}; path=/`;
```
patrick91, line 31: why are you using cookies for the language?

marcoacierno: we can access them on the server and redirect server side

marcoacierno: otherwise we would have to load the JS, flash the UI for a second and then redirect to the correct path

patrick91: ok, just saw it below :) we don't use the user language by default anymore?

marcoacierno: I wanted to add it back in another PR so keep this one simple, I can do it in this one too

patrick91: It’s ok to do in another PR :)

PR #2330, opened 2021-08-05
Use Strawberry's create_type

estyxx:


association-backend/pyproject.toml
```diff
@@ -22,7 +22,7 @@ websockets = "^8.1"
 psycopg2 = "^2.8.6"
 time-machine = "^2.1.0"
 python-dateutil = "^2.8.1"
-pythonit-toolkit = "^0.1.53"
+pythonit-toolkit = "0.1.55-dev.1628202090"
```
marcoacierno, line 25: don't forget to change it back to a normal release

association-backend/pyproject.toml
```diff
@@ -10,7 +10,7 @@ starlette = "^0.14.1"
 uvicorn = "^0.13.3"
 click = "7.1.2"
 invoke = "1.5.0"
-strawberry-graphql = "^0.57.1"
+strawberry-graphql = "^0.70"
```
marcoacierno, line 13: you need to update this on users-backend as well?

estyxx: Yes because was very behind and the `create_type` was release from `0.58.0` so I thought to just upgrade to the last available...

marcoacierno: Yeah sorry I meant that you need to update the other service as well as I don't see the change in users-backend. I am not sure how the tests even pass 🤔  maybe it already has the latest version?

toolkit/pythonit_toolkit/api/extensions.py
```diff
@@ -1,14 +1,15 @@
 from sentry_sdk import add_breadcrumb
 from strawberry.extensions import Extension
+from strawberry.types.execution import ExecutionContext
```
marcoacierno, line 3: what is this change? adding back ExecutionContext?

estyxx: Yeah, think initially I did this change here wrongly, and during the merge git reverted by mistake... 
And surprisingly the tests are working?!?! 🤔

marcoacierno (approved):

PR #2335, opened 2021-08-06
e2e setup

marcoacierno:


frontend/src/components/login-form/index.tsx
```diff
@@ -82,13 +82,19 @@ export const LoginForm: React.SFC<FormProps> = ({ next, ...props }) => {
...
+            variant={message.type}
+            key={message.message}
+          >
             {message.message}
           </Alert>
         ))}
 
         {loginData?.login?.__typename === "WrongEmailOrPassword" && (
-          <Alert variant="alert">Wrong username or password</Alert>
+          <Alert data-testid="wrong-username-or-password-alert" variant="alert">
```
estyxx, line 95: I think I will agree with @marcoacierno  (saw a discussion about in the test setup but was to slow to respond :D) to use data-testid only for inputs/buttons not for messages and alert. 
At least this is how we are testing at work in frontend applications

PR #2338, opened 2021-08-09
PyCon Frontend jest setup

marcoacierno:


frontend/src/components/login-form/index.spec.tsx
```diff
@@ -0,0 +1,272 @@
...
+
+describe("Login form", () => {
+  test("should render", () => {
+    render(
+      <MockedProvider>
+        <LoginForm />
+      </MockedProvider>,
+    );
+
+    expect(screen.getByText("Don't have an account?")).toBeInTheDocument();
```
patrick91, line 74: might be best to use ids :D

patrick91 (approved):

frontend/src/components/login-form/index.spec.tsx
marcoacierno: https://testing-library.com/docs/guiding-principles/

https://testing-library.com/docs/queries/bytestid/

Testing by ID is not recommended as the tests should test what the user sees as far I understood

patrick91: > That said, they are way better than querying based on DOM structure or styling css class names

It's not saying to use text :D I don't really like the idea of hardcoding the labels in the tests :)

marcoacierno: Test ID is the only alternative to match it I think then, because css classes are randomly generated

marcoacierno: > > That said, they are way better than querying based on DOM structure or styling css class names
> 
> It's not saying to use text :D I don't really like the idea of hardcoding the labels in the tests :)

I am also not sure about this, I think querying labels/text is quite common in the FE. I will have a look in a bit and see what others do

PR #2342, opened 2021-08-11
Use a shared library for eslint

marcoacierno:
Use a shared lib instead of having to add eslint dependencies to all JS projects


marcoacierno:
eslint on the gateway is failing, I will fix in another PR

marcoacierno:
Fixed all gateway errors in #2343

association-frontend/package.json
```diff
@@ -35,21 +35,15 @@
     "@graphql-codegen/typescript-operations": "^1.17.14",
     "@graphql-codegen/typescript-react-apollo": "^2.2.1",
     "@graphql-codegen/typescript-urql": "^2.0.3",
+    "@python-italia/eslint-config-pycon": "^1.0.4",
```
patrick91, line 38: nice

patrick91: we could one for all our js apps instead of just pycon maybe? :D

patrick91 (approved):

association-frontend/package.json
marcoacierno: I can name the library "@python-italia/eslint-config-pythonit"

marcoacierno: Created `@python-italia/eslint-config` as global eslint config for python italia projects

Now we just need

```json
{
  "extends": ["@python-italia"]
}
```

to use it

PR #2431, opened 2021-10-29
Internal GraphQL client

estyxx:


toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
+import os
+from typing import Any, Dict, Optional
+
+import httpx
+import jwt
+from pythonit_toolkit.headers import SERVICE_JWT_HEADER
+
+INTERNAL_JWT_SECRET = os.getenv("INTERNAL_JWT_SECRET", "a-very-secret-secret")
```
marcoacierno, line 8: I think the token should be passed when initializating the Client instead of forcing all services to have it as environment variable :)

toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
...
+
+class Client:
+    url = "/internal-api"
+
+    async def execute(
+        self,
+        query: str,
+        variables: Optional[Dict[str, Any]] = None,
+    ):
+        token = jwt.encode({"some": "payload"}, INTERNAL_JWT_SECRET, algorithm="HS256")
```
marcoacierno, line 19: why `some: payload`?

toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
...
+
+        async with httpx.AsyncClient() as client:
+            response = await client.post(
+                self.url,
```
marcoacierno, line 23: uhm how does it work? why only `/internal-api`?

toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
...
+                data={"query": query, "variables": variables},
+                headers={
+                    SERVICE_JWT_HEADER: token,
+                },
+            )
+
+            data = await response.json()
+            return data
+
+    def users(self):
```
marcoacierno, line 33: I don't think the client should have queries in it. It ideally should only expose your `execute` and let the various services send the query

toolkit/pythonit_toolkit/api/client.py
marcoacierno: also this token isn't generated correctly. Ideally it should contain for which service it was generated and what is the destination.

Also each service has it is own internal api, so `url` is not unique

marcoacierno: This is how it is generated on the Gateway: https://github.com/pythonitalia/pycon/blob/8900d9a24ad9b058dee638f206c5dedae183c36d/gateway/pastaporto/user-info.ts#L20

They are also very short-lived, 1 minute max

Instead of having the token generation in the execute method, I would create a method in the `tokens.py` file (if we have it) or at least some different file

toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
...
+INTERNAL_JWT_SECRET = os.getenv("INTERNAL_JWT_SECRET", "a-very-secret-secret")
+
+
+class Client:
+    url = "/internal-api"
+
+    async def execute(
+        self,
+        query: str,
+        variables: Optional[Dict[str, Any]] = None,
```
marcoacierno, line 17: so I am thinking here as input you should accept the service it is calling, the service is it calling from, variables

toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
...
+
+INTERNAL_JWT_SECRET = os.getenv("INTERNAL_JWT_SECRET", "a-very-secret-secret")
+
+
+class Client:
+    url = "/internal-api"
+
+    async def execute(
+        self,
+        query: str,
```
marcoacierno, line 16: let's call it document, this client can be used to execute mutations too

```suggestion
        document: str,
```

users-backend/users/domain/repository.py
```diff
@@ -42,8 +41,12 @@ class UsersRepository(AbstractUsersRepository):
...
         self.session = session
 
-    async def get_users(self) -> Paginable[User]:
-        return Paginable(self.session, User)
+    async def get_users(self, paginate=True) -> Union[list[User], Paginable[User]]:
+        if paginate:
+            return Paginable(self.session, User)
+
+        users = (await self.session.execute(select(User))).scalars().all()
+        return users
```
marcoacierno, line 49: Why are we making a query that returns ALL users? We are talking about 1k or more

users-backend/users/domain/repository.py
```diff
@@ -42,8 +41,12 @@ class UsersRepository(AbstractUsersRepository):
     def __init__(self, session: Optional[AsyncSession] = None) -> None:
         self.session = session
 
-    async def get_users(self) -> Paginable[User]:
-        return Paginable(self.session, User)
+    async def get_users(self, paginate=True) -> Union[list[User], Paginable[User]]:
```
marcoacierno, line 44: can you maybe just implement the query in this PR so we can discuss this change when more in context?

toolkit/tests/client/test_client.py
```diff
@@ -0,0 +1,40 @@
...
+from pythonit_toolkit.api.client import Client
+from ward import test
+
+
+@dataclass
+class MockResponse:
+    data: List[Dict[str, Any]]
+
+    async def json(self):
+        print("take your data!")
```
marcoacierno, line 14:
```suggestion
```

toolkit/pythonit_toolkit/api/client.py
marcoacierno: also when you move it, let's avoid defaults. Ideally it should throw an error.

This is because default might mean that if we ever forget to set this value, we will basically have our secret exposed/known on GH :)

users-backend/users/domain/repository.py
estyxx: because you told me to do so :D

toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
+import os
```
marcoacierno, line 1: I am not sure on the namespace of the client, `api/` I would prefer `service_client` maybe? it needs to have a name that explain what it does

toolkit/pythonit_toolkit/api/client.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import Any, Dict, Optional
+
+import httpx
+import jwt
+from pythonit_toolkit.headers import SERVICE_JWT_HEADER
+
+INTERNAL_JWT_SECRET = os.getenv("INTERNAL_JWT_SECRET", "a-very-secret-secret")
+
+
+class Client:
```
marcoacierno, line 11: Client is a quite generic name 🤔 ServiceClient?

toolkit/pythonit_toolkit/api/client.py
estyxx: We need it only for the internal-api , right?

users-backend/users/domain/repository.py
marcoacierno: I told you that paginating in an internal api doesn’t make much sense

having a query that returns all users, what would be the point?

toolkit/pythonit_toolkit/api/client.py
marcoacierno: The internal api is an api that can exist on any service, so it needs to have the url of the service you are calling

marcoacierno: So association backend might call pycon backend

pycon backend might call users backend

and so on

toolkit/pythonit_toolkit/api/client.py
estyxx: Yeah I add them here in the execute method. So any time we want to call a query we have to include issuer and audience. I thought that maybe in the future can be the case where we want to use the same client to call different services...

toolkit/pythonit_toolkit/api/client.py
estyxx: Yeah I know is not ideal to return all the attributes any time. I thought we can maybe return an object instead of a dictionary if we know what we are looking for...

toolkit/pythonit_toolkit/api/client.py
estyxx: So I add the url in the init method

users-backend/users/domain/repository.py
estyxx: Yeah, probably we don't need all the user... I guess I can pass a list of ids

users-backend/users/domain/repository.py
```diff
@@ -43,7 +43,8 @@ def __init__(self, session: Optional[AsyncSession] = None) -> None:
         self.session = session
 
     async def get_users(self) -> Paginable[User]:
-        return Paginable(self.session, User)
+        users = (await self.session.execute(select(User))).scalars().all()
+        return users
```
marcoacierno, line 47: let's just implement the client in this API

toolkit/tests/service_client/test_client.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    mock_data = [
+        {
+            "id": 1,
+        },
+    ]
+
+    with patch("httpx.AsyncClient.post") as post_mock:
+        post_mock.return_value = MockResponse(mock_data)
+        cleint = ServiceClient(
```
marcoacierno, line 34:
```suggestion
        client = ServiceClient(
```

toolkit/pythonit_toolkit/api/service_client.py
```diff
@@ -0,0 +1,40 @@
...
+            response = await client.post(
+                self.url,
+                data={"query": document, "variables": variables},
+                headers={
+                    SERVICE_JWT_HEADER: token,
+                },
+            )
+
+            data = await response.json()
+            return data
```
marcoacierno, line 40: maybe we can parse this and return a dataclass

```
class ServiceResponse:
    errors: Optional[List]
    data: Optional[str]
```

so in the future we could play a lot more with it and do stuff like `.has_errors` and so on

toolkit/tests/service_client/test_client.py
```diff
@@ -0,0 +1,43 @@
...
+            url="http://localhost:8050",
+            issuer="pycon",
+            audience="users-service",
+            jwt_secret="mysecret",
+        )
+
+        response = await cleint.execute(document=query)
+
+        assert response == mock_data
```
marcoacierno, line 43: I don't think this test is correct, the client doesn't respond with the data directly

marcoacierno (changes requested):
left some comments

toolkit/pythonit_toolkit/api/service_client.py
```diff
@@ -0,0 +1,47 @@
...
+        self,
+        url: str,
+        issuer: str,
+        audience: str,
+        jwt_secret: str,
+    ):
+        self.url = url
+        self.issuer = issuer
+        self.audience = audience
+        self.jwt_secret = jwt_secret
```
marcoacierno, line 26: we should raise an error if those are empty

toolkit/pythonit_toolkit/api/service_client.py
```diff
@@ -0,0 +1,47 @@
...
+            response = await client.post(
+                self.url,
+                data={"query": document, "variables": variables},
+                headers={
+                    SERVICE_JWT_HEADER: token,
+                },
+            )
+
+            data = await response.json()
+            return ServiceResponse(errors=data.get("errors"), data=data.get("data"))
```
marcoacierno, line 47: what happens if the server response is a 400?

I would add a `response.raise_for_status()` so at least we throw an exception if the response isn't a 200

In the future I imagine here we will have automatic retry, or implement a service mesh on AWS :P

toolkit/pythonit_toolkit/pastaporto/tokens.py
```diff
@@ -24,3 +24,15 @@ def decode_service_to_service_token(
         algorithms=["HS256"],
         options={"require": ["exp", "iss", "aud", "iat"]},
     )
+
+
+def generate_token(secret, issuer: str, audience: str, expires_in: str = "1m"):
```
marcoacierno, line 29: can we add some unit-tests for this?

toolkit/pythonit_toolkit/pastaporto/tokens.py
```diff
@@ -24,3 +24,15 @@ def decode_service_to_service_token(
...
+    return jwt.encode(
+        {
+            "issuer": issuer,
+            "audience": audience,
+            "expires_in": expires_in,
```
marcoacierno, line 34: where is this `expires_in` coming from? jwt payload expects `exp` for the expiration

PyJWT docs: https://pyjwt.readthedocs.io/en/latest/usage.html#expiration-time-claim-exp

e.g https://github.com/pythonitalia/pycon/blob/be4ccaeee6253b0e66c1106d5f2e59291696c5fe/users-backend/users/domain/entities.py#L77

marcoacierno: also it can't be "1m", it has to be a datetime, from the documentation:

```
datetime.datetime.now(tz=timezone.utc) + datetime.timedelta(seconds=30)
```

marcoacierno (commented):
left some more comments

toolkit/tests/service_client/test_client.py
```diff
@@ -0,0 +1,39 @@
...
+from pythonit_toolkit.api.service_client import ServiceClient
+from ward import test
+
+
+@dataclass
+class MockResponse:
+    return_value: Optional[Any] = None
+
+    async def json(self):
+        return self.return_value
```
marcoacierno, line 14: I don't think you need to do this

toolkit/tests/service_client/test_client.py
```diff
@@ -0,0 +1,39 @@
...
+        query{
+            users {
+                id
+            }
+        }
+    """
+    mock_response = {"data": {"users": [{"id": 1}]}}
+
+    with patch("httpx.AsyncClient.post") as post_mock:
+        post_mock.return_value = MockResponse(return_value=mock_response)
```
marcoacierno, line 29: something like this should work so you don't need the MockResponse dataclass? mocks usually are very flexible

```suggestion
        from unittest.mock import Mock
        post_mock.return_value = Mock()
        post_mock.return_value.json.return_value = mock_response
```

marcoacierno (commented):
left some minor comments about the mock

toolkit/pythonit_toolkit/pastaporto/tokens.py
marcoacierno: we should have a more specific name for this function

also expiration should always be 1 minute so let's not pass it as parameter

```suggestion
def generate_service_to_service_token(secret, issuer: str, audience: str):
```

toolkit/pythonit_toolkit/pastaporto/tokens.py
```diff
@@ -24,3 +24,15 @@ def decode_service_to_service_token(
         algorithms=["HS256"],
         options={"require": ["exp", "iss", "aud", "iat"]},
     )
+
+
+def generate_token(secret, issuer: str, audience: str, expires_in: str = "1m"):
+    return jwt.encode(
+        {
+            "issuer": issuer,
+            "audience": audience,
```
marcoacierno, line 33: Uhm I think what you want here is `iss` and `aud`: https://pyjwt.readthedocs.io/en/latest/usage.html#issuer-claim-iss

toolkit/pythonit_toolkit/pastaporto/tokens.py
marcoacierno: also type hint missing in secret :)

```suggestion
def generate_token(secret: str, issuer: str, audience: str, expires_in: str = "1m"):
```

can we also raise a ValueError `if not secret`? same for issuer and audience, those should always be there and not be empty

toolkit/pythonit_toolkit/api/service_client.py
```diff
@@ -0,0 +1,47 @@
...
+    errors: Optional[Dict[str, Any]]
+    data: Optional[Dict[str, Any]]
+
+
+class ServiceClient:
+    def __init__(
+        self,
+        url: str,
+        issuer: str,
+        audience: str,
```
marcoacierno, line 20: we can keep it as it is for now, but I think we should make `url` and `audience` simpler, we are contacting a service so this __init__ could technically just take a `service_name` as property, and find the URL in some other way? what do you think? I just need to think of a way to have the list of URLs, we could take them from env variables but I am not a fan of forcing env variables at the toolkit level

marcoacierno (commented):
left some comments

toolkit/pythonit_toolkit/pastaporto/tokens.py
```diff
@@ -24,3 +24,20 @@ def decode_service_to_service_token(
         algorithms=["HS256"],
         options={"require": ["exp", "iss", "aud", "iat"]},
     )
+
+
+def generate_service_to_service_token(
+    secret, issuer: str, audience: str, expires_in: str = "1m"
```
marcoacierno, line 30: let's not accept the expiration as param, we don't really want long duration tokens

toolkit/tests/service_client/test_client.py
estyxx: but json is async, that is why I made a custom class... I wasn't able to make it work with a normal Mock object... :(

toolkit/pythonit_toolkit/api/service_client.py
estyxx: Agree

toolkit/tests/service_client/test_client.py
estyxx: Uhh there is also AsyncMock!

toolkit/tests/service_client/test_client.py
```diff
@@ -0,0 +1,87 @@
...
+        "users-service",
+        "",
+        "Argument 'jwt_secret' can't be empty",
+    ),
+]:
+
+    @test("raise ValueError if arguments are empty")
+    async def _():
+
+        with raises(ValueError) as exc:
```
marcoacierno, line 79: I don't know what to think about this :D does it even work?

estyxx: It's how ward suggests doing parametrised tests... https://ward.readthedocs.io/en/latest/guide/writing_tests.html#parameterised-testing

marcoacierno: They also have `each`: https://github.com/darrenburns/ward/blob/a84a4807f8b12dd1fc38da05814031095d62f6bf/tests/test_rewrite.py#L70

Maybe it will look better?

marcoacierno: Anyway separate test might even work for me, I know it is duplicate, but it would much clearer

toolkit/pythonit_toolkit/api/service_client.py
```diff
@@ -0,0 +1,53 @@
...
+                data={"query": document, "variables": variables},
+                headers={
+                    SERVICE_JWT_HEADER: token,
+                },
+            )
+
+            response.raise_for_status()
+
+            data = await response.json()
+            return ServiceResponse(errors=data.get("errors"), data=data.get("data"))
```
marcoacierno, line 53: What do you think about something like this?

```suggestion
            response.raise_for_status()

            data = await response.json()
            errors = data.get("errors", None)
            
            if errors:
                return ServiceError(errors)
                
            return ServiceResponse(data=data.get("data"))
```

where `ServiceError` is

```python
@dataclass
class ServiceError(Exception):
    errors: Dict[str, Any]
```

`ServiceResponse` will only have `data`.

Why I am thinking this? This would allow us to do

```python
try:
    client.execute()
expect ServiceError:
    # error validation
```

instead of having to do `if`s in the execute block. What  do you think? would it look better?

toolkit/pythonit_toolkit/api/service_client.py
```diff
@@ -0,0 +1,53 @@
...
+        self.url = url
+        self.issuer = issuer
+        self.audience = audience
+        self.jwt_secret = jwt_secret
+
+        for arg in ("url", "issuer", "audience", "jwt_secret"):
+            if not getattr(self, arg):
+                raise ValueError(f"Argument '{arg}' can't be empty")
+
+    async def execute(
```
marcoacierno, line 32: can we use async in the pycon backend since it is django? 🤔

marcoacierno: actually maybe it doesn't matter if it is django or not, but how we run it on the lambda, if we run the ASGI version? I don't remember what was needed to run django under to have async working 🤔

toolkit/tests/service_client/test_client.py
estyxx: Four test for some stupid missing arguments?!? it's not too much? 
I changed to use each btw...

toolkit/pythonit_toolkit/api/service_client.py
marcoacierno:
```suggestion
        service_name: str,
        url: str,
        caller: str,
```

service_name is the destination (audience in the JWT)
url is the URL of the destination's service
caller is the service calling the destination (issuer in the JWT)

I would use those names in the service client, since the names of the JWT properties do not make sense in the client context

toolkit/pythonit_toolkit/api/service_client.py
estyxx: Nice and cleaner! :)

toolkit/tests/service_client/test_client.py
marcoacierno: You are writing tests, not business logic 🙏 tests should be clear and easy to read.

> "Four test for some stupid missing arguments"

You are not just testing missing arguments, you are testing that your client interface is correct so users of the library don't do anything weird with it/use it wrong

toolkit/pythonit_toolkit/pastaporto/tokens.py
```diff
@@ -24,3 +25,18 @@ def decode_service_to_service_token(
         algorithms=["HS256"],
         options={"require": ["exp", "iss", "aud", "iat"]},
     )
+
+
+def generate_service_to_service_token(secret, issuer: str, audience: str):
```
marcoacierno, line 30:
```suggestion
def generate_service_to_service_token(secret: str, issuer: str, audience: str):
```

toolkit/pythonit_toolkit/pastaporto/tokens.py
```diff
@@ -24,3 +25,18 @@ def decode_service_to_service_token(
...
+    if not secret:
+        raise ValueError("Secret can not be empty")
+
+    return jwt.encode(
+        {
+            "iss": issuer,
+            "aud": audience,
+            "exp": datetime.now(tz=timezone.utc) + timedelta(seconds=30),
```
marcoacierno, line 38: uhm can we do 60 seconds?

toolkit/pythonit_toolkit/pastaporto/tokens.py
```diff
@@ -24,3 +25,18 @@ def decode_service_to_service_token(
...
+        },
+        secret,
```
marcoacierno, line 40:
```suggestion
        str(secret),
```

it is not needed, but makes our life easier when in starlette we use `Config#Secret`

toolkit/pythonit_toolkit/service_client/client.py
```diff
@@ -0,0 +1,61 @@
...
+                    SERVICE_JWT_HEADER: token,
+                },
+            )
+
+            response.raise_for_status()
+            data = await response.json()
+            errors = data.get("errors", None)
+
+            if errors:
+                return ServiceError(errors)
```
marcoacierno, line 59:
```suggestion
                raise ServiceError(errors)
```

toolkit/tests/service_client/test_client.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+        response = await client.execute(document="""{ users { id }}""")
+
+        assert response.errors == ["Something went wrong"]
+
+
+@test("raise ValueError if arguments are empty")
+async def _(
+    url=each(
```
marcoacierno, line 55: this just looks terrible :(

toolkit/tests/service_client/test_client.py
```diff
@@ -0,0 +1,95 @@
...
+            audience="users-service",
+            jwt_secret="mysecret",
+        )
+
+        response = await client.execute(document=query)
+
+        assert response.data == {"users": [{"id": 1}]}
+
+
+@test("raise Exception")
```
marcoacierno, line 34: this should explain what the test is testing

toolkit/tests/pastaporto/test_tokens.py
```diff
@@ -72,3 +75,19 @@ async def _():
...
             test_token, "secret", issuer="test", audience="association-backend"
         )
+
+
+@test("generate a token")
+async def _():
+    token = generate_service_to_service_token("secret", issuer="me", audience="you")
+
+    assert token
+
```
marcoacierno, line 85: This test doesn't seem very useful, should we test it against a known token? Maybe freeze the time with time_machine

toolkit/pythonit_toolkit/service_client/client.py
```diff
@@ -0,0 +1,61 @@
...
+    errors: Dict[str, Any]
+
+
+class ServiceClient:
+    def __init__(
+        self,
+        url: str,
+        issuer: str,
+        audience: str,
+        jwt_secret: str,
```
marcoacierno, line 25: can you do the changes I suggested here: https://github.com/pythonitalia/pycon/pull/2431#discussion_r748662345?

toolkit/tests/pastaporto/test_tokens.py
```diff
@@ -72,3 +75,19 @@ async def _():
         decode_service_to_service_token(
             test_token, "secret", issuer="test", audience="association-backend"
         )
+
+
+@test("generate a token")
```
marcoacierno, line 80: this should describe what the test is doing

```suggestion
@test("Generate a service to service token")
```

toolkit/tests/pastaporto/test_tokens.py
```diff
@@ -72,3 +75,19 @@ async def _():
...
+async def _():
+    token = generate_service_to_service_token("secret", issuer="me", audience="you")
+
+    assert token
+
+
+@test("raise ValueError if secret is empty")
```
marcoacierno, line 87: this file contains multiple tests for different types of tokens, maybe the description should be more clear?

```suggestion
@test("Secret is required when creating a service-to-service token")
```

marcoacierno:
I fixed a few errors and now the client works 🥳 

Example in django:

```python
client = ServiceClient(
    url='http://users-backend:8050/internal-api',
    service_name='users-backend',
    caller='pycon-backend',
    jwt_secret='service-to-service-secret'
)
client_execute = async_to_sync(client.execute)

(Pdb++) client_execute("{__typename}")
ServiceResponse(data={'__typename': 'Query'})
```

The users internal API had some errors/assumption that it would be only called from the gateway, so I prepared another PR to fix those: https://github.com/pythonitalia/pycon/pull/2470/files

marcoacierno:
Some tests are failing, we can have a look tomorrow. Thanks for this!

marcoacierno (approved):
Have a look at the last changes I did, but I am happy to merge this :) 

@estyxx one thing, this PR has 40+ commits which we don't really need, when you merge can you select the "squash" option?

PR #2467, opened 2021-11-10
Call mailchimp api

estyxx:
CleanShot 2021-11-10 at 22 21 30
CleanShot 2021-11-10 at 22 21 04

deploy notes:

add to AWS secrets both for staging&prod:

MAILCHIMP_LIST_ID
MAILCHIMP_SECRET_KEY


backend/api/newsletters/mailchimp.py
```diff
@@ -0,0 +1,24 @@
+from dataclasses import dataclass
```
marcoacierno, line 1: this file is inside `api/` which doesn't seem right, in `api` we should only have GraphQL related stuff, this is more of an integration

We already have a `integrations` namespace in the pycon backend source, so we could move this file there?

backend/pycon/settings/base.py
```diff
@@ -202,3 +202,8 @@
 QUERY_INSPECT_TRACEBACK_ROOTS = [root(".")]
 
 PINPOINT_APPLICATION_ID = env("PINPOINT_APPLICATION_ID", default="")
+
+
+MAILCHIMP_SECRET_KEY = env("MAILCHIMP_SECRET_KEY")
+MAILCHIMP_DC = "us5"
+MAILCHIMP_LIST_ID = env("MAILCHIMP_LIST_ID")
```
marcoacierno, line 209: I think we can make those optional and in the `subscribe` of mailchimp if `MAILCHIMP_SECRET_KEY` is None raise an error saying `Mailchimp integration is not enabled`

What do you think?

backend/api/newsletters/mailchimp.py
```diff
@@ -0,0 +1,24 @@
...
+        "Accept": "application/json",
+        "Authorization": f"Bearer {settings.MAILCHIMP_SECRET_KEY}",
+    }
+    response = requests.post(
+        url, json={"email_address": email, "status": "subscribed"}, headers=headers
+    )
+
+    data = response.json()
+
+    return MailchimpSubscription(id=data["id"], email=data["email_address"])
```
marcoacierno, line 24: we should have error handling here, what if Mailchimp raises an error? we need to catch it and raise a custom exception so that we can return an error to GraphQL saying "Something went wrong, please try again"

backend/api/newsletters/forms.py
```diff
@@ -1,16 +1,16 @@
 from django import forms
 from newsletters.models import Subscription
 from strawberry_forms.forms import FormWithContext
+from api.newsletters.mailchimp import subscribe
 
 
 class SubscribeToNewsletterForm(FormWithContext):
```
marcoacierno, line 7: can we add some unit-tests for this mutation?

backend/pycon/settings/base.py
```diff
@@ -202,3 +202,8 @@
 QUERY_INSPECT_TRACEBACK_ROOTS = [root(".")]
 
 PINPOINT_APPLICATION_ID = env("PINPOINT_APPLICATION_ID", default="")
+
+
+MAILCHIMP_SECRET_KEY = env("MAILCHIMP_SECRET_KEY")
+MAILCHIMP_DC = "us5"
```
marcoacierno, line 208: tomorrow morning I will check what is our DC value, if it is us5 too we could just not add it to settings/base.py and just have it hardcoded in the URL but it is the same, it can stay here too

backend/api/newsletters/forms.py
```diff
@@ -1,16 +1,16 @@
...
     email = forms.EmailField()
 
     def save(self):
         email = self.cleaned_data.get("email")
-        subscription, _ = Subscription.objects.get_or_create(email=email)
 
-        return subscription
+        return subscribe(email)
```
marcoacierno, line 13: what is sent back to the FE?

marcoacierno (commented):
left some comments :)

backend/pycon/settings/base.py
estyxx: raise something directly in the settings file? or only in prod file?

backend/api/newsletters/forms.py
```diff
@@ -1,16 +1,25 @@
...
 
     def save(self):
         email = self.cleaned_data.get("email")
-        subscription, _ = Subscription.objects.get_or_create(email=email)
 
-        return subscription
+        try:
+            return subscribe(email)
+        except Exception as e:
+            logger.error(e, exc_info=True)
```
marcoacierno, line 21: also `except Exception as e` feels very generic 🤔 

```suggestion
            logger.error("Unable to subscribe the user to mailchimp due to an error", e, exc_info=True)
```

backend/api/newsletters/mutations.py
```diff
@@ -1,17 +1,19 @@
-from api.newsletters.forms import SubscribeToNewsletterForm, UnsubscribeToNewsletterForm
-from api.newsletters.types import NewsletterSubscription
+from api.newsletters.forms import (
+    SussbscribeToNewsletterForm,
```
marcoacierno, line 2: `Suss` :sus:

backend/api/tests/schema/test_newsletter.py
```diff
@@ -3,11 +3,11 @@
 
 from newsletters.models import Subscription
 
-pytestmark = mark.skip
+# pytestmark = mark.skip
```
marcoacierno, line 6:
```suggestion
```

backend/api/newsletters/mutations.py
estyxx: lol

backend/api/tests/schema/test_newsletter.py
```diff
@@ -3,11 +3,11 @@
...
 
 
-@mark.django_db
-def test_subscribe_to_newsletter(graphql_client):
+def test_subscribe_to_newsletter(graphql_client, mocker):
```
estyxx, line 8: I don't know why this test is failing: 

I have patched the method to return True, but seems like is calling the real API... I'm confused... 🤔 


```python
api/tests/schema/test_newsletter.py:31: AssertionError
------------------------------ Captured log call -------------------------------
ERROR    integrations.mailchimp:mailchimp.py:34 Error from mailchimp: {'type': 'mailchimp.com/developer/marketing/docs/errors', 'title': 'Forbidden', 'status': 403, 'detail': "The API key provided is linked to datacenter 'secret'", 'instance': 'e141fa98-4f5f-0e1a-e632-a064ab5a6535'}
ERROR    api.newsletters.forms:forms.py:21 403 Forbidden: The API key provided is linked to datacenter 'secret'
Traceback (most recent call last):
  File "/home/runner/work/pycon/pycon/backend/api/newsletters/forms.py", line 19, in save
    return subscribe(email)
  File "/home/runner/work/pycon/pycon/backend/integrations/mailchimp.py", line 35, in subscribe
    raise MailchimpError(
integrations.mailchimp.MailchimpError: 403 Forbidden: The API key provided is linked to datacenter 'secret'
```

backend/integrations/mailchimp.py
```diff
@@ -0,0 +1,39 @@
...
+    data = response.json()
+    if data["status"] == "subscribed" or data["title"] == "Member Exists":
+        return True
+    else:
+        logger.error("Error from mailchimp: %s", data)
+        raise MailchimpError(
+            status=data.get("status"),
+            title=data.get("title"),
+            detail=data.get("detail"),
+        )
```
marcoacierno, line 39: no need for the else

```suggestion
    if data["status"] == "subscribed" or data["title"] == "Member Exists":
        return True

    logger.error("Error from mailchimp: %s", data)
    raise MailchimpError(
        status=data.get("status"),
        title=data.get("title"),
        detail=data.get("detail"),
    )
```

backend/pycon/settings/base.py
```diff
@@ -195,7 +195,12 @@
...
     "handlers": {"console": {"class": "logging.StreamHandler"}},
     "loggers": {
-        "app_api": {"handlers": ["console"], "level": "WARNING"},
+        "pycon.api": {"handlers": ["console"], "level": "WARNING", "propagate": True},
+        "pycon.integrations": {
+            "handlers": ["console"],
+            "level": "WARNING",
+            "propagate": True,
+        },
         "qinspect": {"handlers": ["console"], "level": "DEBUG", "propagate": True},
```
marcoacierno, line 204: does this work the same?

```suggestion
        "pycon": {"handlers": ["console"], "level": "WARNING", "propagate": 
        "qinspect": {"handlers": ["console"], "level": "DEBUG", "propagate": True},
```

docker-compose.yml
```diff
@@ -11,13 +11,16 @@ x-defaults:
     API_URL_SERVER: http://gateway:4000/graphql
     # Stripe
     STRIPE_SUBSCRIPTION_PRICE_ID: price_1IkVzxD5MZ3GejSORRBZCvK6
+    # Mailchimp
+    MAILCHIMP_LIST_ID: "12345abcdf89"
```
marcoacierno, line 15:
```suggestion
```

docker-compose.yml
```diff
@@ -11,13 +11,16 @@ x-defaults:
...
     # URLs
     ASSOCIATION_FRONTEND_URL: http://localhost:3020
     # Secrets
     PASTAPORTO_SECRET: pastaporto-secret
     IDENTITY_SECRET: identity-secret
     SERVICE_TO_SERVICE_SECRET: service-to-service-secret
     PASTAPORTO_ACTION_SECRET: pastaporto-action-secret
+    MAILCHIMP_SECRET_KEY: mailchimp-secret
```
marcoacierno, line 23:
```suggestion
```

marcoacierno: for local development we could do something similar to what we did for `STRIPE_SECRET_API_KEY` for example, so if you want to enable the integration, you have to edit your `.env` file

backend/api/tests/schema/test_newsletter.py
marcoacierno: I think you have to do

```python
with mocker.patch("integrations.mailchimp.subscribe") as mock_subscription:
    mock_subscription.return_value = True
    # call the API inside the with block
```

otherwise right now it doesn't patch anything

backend/pycon/settings/base.py
estyxx: think before we were missing some logs, `app_api` was wrong. 
If you look at [Django logging docs](https://docs.djangoproject.com/en/3.2/topics/logging/#configuring-logging) the correct configuration is ` 'myproject.custom'`. 
And also `propagate` means you configure the log to make it work also in the submodules.

docker-compose.yml
estyxx: yeah but there are defaults for the other secrets, why remove this?

backend/api/tests/schema/test_newsletter.py
estyxx: I tried also that but was not changing so much... I will give another shot...

estyxx: I also tested on docker locally and the test was working now... but in the CI it fails again... :(  
I don't understand what it's happening 🤔

docker-compose.yml
marcoacierno: the other secrets (SERVICE_TO_SERVICE_SECRET etc) have a meaning, for example the one I sent here it is the shared secret that services use to communicate.

`MAILCHIMP_SECRET_KEY: mailchimp-secret` means nothing, it will only make the pycon backend subscription code fail contacting mailchimp saying "the api key is not valid", 

that's the reason I told you to add this check

```python
    if not settings.MAILCHIMP_SECRET_KEY:
        raise ValueError("Mailchimp integration is not configured")
```

it is because it is a better DX error and makes clear to who sees it that they need to configure mailchimp if they want to test it.

If you add random values they would get random errors from mailchimp

marcoacierno: so what I am suggesting here is to do the same as how we configurated stripe (see there are no defaults for stripe?) so that it gets the value from .env

This allows a person to go to mailchimp and prepare the secret key, list id and so on so they can test the mailchimp integration and add their own values in .env

marcoacierno:
@estyxx do you have to update the infrastructure terraform files to pass those 2 keys as env variables or do you prefer I do it?

I will update AWS Parameter store in a bit too to have those values

docker-compose.yml
estyxx: okay

estyxx: but there is the STRIPE_SUBSCRIPTION_PRICE_ID , so should I left the LIst Id for Mailchimp??

marcoacierno (approved):
Happy to merge this :)

PR #2470, opened 2021-11-14
Fix internal api checks

marcoacierno:


users-backend/users/internal_api/permissions.py
```diff
@@ -1,23 +1,39 @@
...
-        jwt.InvalidIssuerError,
-        jwt.ExpiredSignatureError,
-        jwt.InvalidAudienceError,
-    ):
-        return False
+def IsService(allowed_callers: list[str]):
+    if not allowed_callers:
+        raise ValueError("No callers allowed specified")
+
+    class _IsService(BasePermission):
```
marcoacierno, line 16: This is worth moving into toolkit + write tests for it

users-backend/users/internal_api/permissions.py
```diff
@@ -1,23 +1,39 @@
...
-        )
-        return True
-    except (
-        jwt.DecodeError,
-        jwt.InvalidIssuerError,
-        jwt.ExpiredSignatureError,
-        jwt.InvalidAudienceError,
-    ):
-        return False
+def IsService(allowed_callers: list[str]):
```
marcoacierno, line 12: @patrick91 Strawberry permissions can't accept parameters, so I am forced to do this like this, I wonder if it might be an improvement we can do to Strawberry to instead of accepting a list of classes as `permission_classes` it accepts the instances 🤔

users-backend/users/internal_api/permissions.py
marcoacierno: I will move it into toolkit once #2431 is merged

users-backend/users/internal_api/tests/test_views.py
```diff
@@ -1,30 +0,0 @@
-from pythonit_toolkit.headers import SERVICE_JWT_HEADER
```
marcoacierno, line 1: removed those tests for now... for sure I still want to add a generic validation that you can't call internal-api without the token, so they hopefully will be back

PR #2472, opened 2021-11-15
Add new permission for internal api service checks

marcoacierno:


toolkit/pythonit_toolkit/api/permissions.py
```diff
@@ -1,9 +1,52 @@
...
+
+    if not secret:
+        raise ValueError("JWT secret cannot be empty")
+
+    if not service:
+        raise ValueError("Current service name cannot be empty")
+
+    secret = str(secret)
+
+    class _IsService(BasePermission):
```
estyxx, line 30: why redefining the class here?

toolkit/tests/api/test_permissions.py
```diff
@@ -0,0 +1,88 @@
...
+from pythonit_toolkit.api.permissions import IsService
+from pythonit_toolkit.headers import SERVICE_JWT_HEADER
+from pythonit_toolkit.pastaporto.tokens import generate_service_to_service_token
+from ward import raises, test
+
+
+@test("Allowed callers pass the permission")
+async def _():
+    test_token = generate_service_to_service_token("test", "gateway", "users-backend")
+    PermissionClass = IsService(["gateway"], "test", "users-backend")
```
estyxx, line 12: can you separate the Arrange-Act-Assert phases with a blank line?

toolkit/tests/api/test_permissions.py
```diff
@@ -0,0 +1,88 @@
...
+    headers = {SERVICE_JWT_HEADER: test_token}
+    mock_info = MagicMock()
+    mock_info.context.request.headers.get = headers.get
+    assert PermissionClass().has_permission(None, mock_info) is True
+
+
+@test("Not allowed allowed callers fail the permission")
```
estyxx, line 19: Is worth adding a test with a schema Query with this permission?
So an "integration" test to see how this permission can be used in schema...

toolkit/pythonit_toolkit/api/permissions.py
marcoacierno: every time the function is called a new class is created, this is because in strawberry you can't have a permission with parameters in the constructor

toolkit/tests/api/test_permissions.py
marcoacierno: cleaned up the tests a bit, let me know

toolkit/tests/api/test_permissions.py
marcoacierno: Uhm, I like the idea but we don't have any test like this in the toolkit so I am not sure how hard it would be to setup 🤔

toolkit/pythonit_toolkit/api/permissions.py
marcoacierno: The `IsService` that you import is actually a function, this allows us to send parameters

marcoacierno: it will then be used like this: https://github.com/pythonitalia/pycon/pull/2470/files#diff-f484eb8ffccc13ba354f92786a410bb34829668243880b0d0e9d1b6909f2f163R16

marcoacierno: here: https://github.com/pythonitalia/pycon/pull/2473

PR #2474, opened 2021-11-16
Update django and add banner with env name

marcoacierno:
Add a banner so we know in which environment we are in when navigating the admin.

image

image

This PR also updates Django to 3.2

### DIFF
--- a/backend/pycon/settings/base.py
+++ b/backend/pycon/settings/base.py
@@ -20,6 +20,7 @@ FRONTEND_URL = env("FRONTEND_URL")
 # Application definition
 
 INSTALLED_APPS = [
+    "custom_admin",
     "dal",
     "dal_select2",
     "dal_admin_filters",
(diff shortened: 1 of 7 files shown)


patrick91 (approved):

estyxx (approved):

backend/pycon/settings/base.py
```diff
@@ -76,15 +77,17 @@
...
         "OPTIONS": {
             "context_processors": [
                 "django.template.context_processors.debug",
                 "django.template.context_processors.request",
                 "django.contrib.auth.context_processors.auth",
                 "django.contrib.messages.context_processors.messages",
-            ]
+                "custom_admin.context_processors.admin_settings",
+            ],
+            "debug": True,
```
pauloxnet, line 90: Maybe it will better to use `DEBUG` environment variable here instead of a `True` as default ?
Something like that:
```suggestion
            "debug": env("DEBUG", default=False)
```

marcoacierno: The debug here doesn't seem to do anything useful so I will just remove it

PR #2475, opened 2021-11-16
Send CFP notifications via SQS

marcoacierno:
## Why

We want to send slack notifications when someone sends a CFP

## How to test

- Go to https://pycon-gvxzz6zpg-python-italia.vercel.app/en/cfp
- Create an account on staging
- Submit a CFP!
- Go to slack and open the channel website-test

You should see your submission there

image


infrastructure/global/iam_roles/lambda_roles.tf
```diff
@@ -65,4 +65,16 @@ data "aws_iam_policy_document" "lambda_role" {
...
+  statement {
+    actions = [
+      "sqs:SendMessage",
+      "sqs:DeleteMessage",
+      "sqs:GetQueueAttributes",
+      "sqs:ReceiveMessage",
+    ]
+    resources = ["*"]
+    effect    = "Allow"
+  }
```
marcoacierno, line 79: Having 1 lambda role for all our lambda makes it simpler to manage but it is impossible to apply specific permissions

i.e right now staging can read/delete messages from the production queue which is not ideal.

we should refactor this sooner than later

backend/wsgi_handler.py
```diff
@@ -35,10 +36,15 @@ def import_app(config):
 
 
 def handler(event, context):
+    if "Records" in event:
+        process_sqs_messages(event)
+        return
```
marcoacierno, line 41: I wanted to use a separate lambda but it simply wasn't worth the mess to do it

backend/api/submissions/forms.py
```diff
@@ -119,12 +119,13 @@ def save(self, commit=True):
...
-            instance.elevator_pitch,
-            instance.type.name,
-            request.build_absolute_uri(instance.get_admin_url()),
-            instance.duration.duration,
-            instance.topic.name,
+            submission_id=instance.id,
+            title=instance.title,
+            elevator_pitch=instance.elevator_pitch,
+            submission_type=instance.type.name,
+            admin_url=request.build_absolute_uri(instance.get_admin_url()),
```
marcoacierno, line 126: this returns the API gateway link, will fix it in another PR :(

marcoacierno, line 126: this returns the wrong URL on staging, but on production should be correct

(this will be fixed once we fix the hack we have to force pycon.it)

infrastructure/global/iam_roles/lambda_roles.tf
patrick91: can make an issue for that? :)

backend/sqs_messages.py
```diff
@@ -0,0 +1,66 @@
...
+            message_id,
+            message_type,
+        )
+        return
+
+    try:
+        data = json.loads(record["body"])
+        handler(data)
+    except Exception as exc:
+        # In future we should re-schedule the message with a delay if it fails
```
estyxx, line 47: At work I saw we use `tenacity.Retrying` https://tenacity.readthedocs.io/en/latest/

marcoacierno: do they support SQS? might be worth having a look in the future

estyxx: Well, this will retry executing the handler whenever an Exception occurs, it will not reschedule the task in SQS.
In the docs it's used only with the decorator, I saw used as a context manager as well, think it's the same...

marcoacierno: uhm ok, my idea was to return them to the queue and process them after a few hours (e.g if slack is down) but it highly depends on what the task is doing so I can see why it retries in-place

PR #2477, opened 2021-11-17
Users widget for django admin

marcoacierno:
Linear issue with all the places where it should be used: https://linear.app/python-italia/issue/WEB-147/django-admin-microservice-utenti

# Why

Django admin needs to display more information about our users, just showing the id is not useful

image

image

image

# How to test

Start docker setup
Go to localhost:3000/
Create some users

Go to the admin blog post http://localhost:8000/admin/blog/post/
open an existing blog post / create a new one

"Author" will now be an autocomplete field

Search the users you created it, it should match

Go over submission, schedule item and others they should all work


backend/submissions/admin.py
```diff
@@ -47,6 +88,17 @@ class SubmissionAdmin(admin.ModelAdmin):
...
+    def speaker_display_name(self, obj):
+        return self._users_by_id[str(obj.speaker_id)]["displayName"]
+
+    speaker_display_name.short_description = "Speaker"
+
+    def get_queryset(self, request):
+        queryset = super().get_queryset(request)
+        author_ids = queryset.values_list("speaker_id", flat=True)
+        self._users_by_id = get_users_data_by_ids(list(author_ids))
+        return queryset
```
patrick91, line 100: should we have a mixin for this?

backend/users/admin_views.py
```diff
@@ -0,0 +1,35 @@
...
+
+def users_autocomplete(request):
+    term = request.GET.get("term", "")
+    client = ServiceClient(
+        url=f"{settings.USERS_SERVICE}/internal-api",
+        service_name="users-backend",
+        caller="pycon-backend",
+        jwt_secret=settings.SERVICE_TO_SERVICE_SECRET,
+    )
+    client_execute = async_to_sync(client.execute)
```
patrick91, line 23: nit: we can add `client.execute_sync` to make this slightly nicer (for another PR anyway)

backend/users/admin_views.py
```diff
@@ -0,0 +1,35 @@
...
+    search_result = client_execute(SEARCH_USERS, {"query": term})
+    search_data = search_result.data
```
patrick91, line 25: should we handle errors?

backend/users/autocomplete.py
```diff
@@ -0,0 +1,69 @@
...
+from users.client import get_users_data_by_ids
+
+
+class UsersBackendAutocomplete(AutocompleteSelect):
+    def __init__(self, admin_site, attrs=None, choices=(), using=None):
+        self.admin_site = admin_site
+        self.db = using
+        self.choices = choices
+        self.attrs = {} if attrs is None else attrs.copy()
+        self.i18n_name = SELECT2_TRANSLATIONS.get(get_language())
```
patrick91, line 16: does this init do anything custom?

backend/users/autocomplete.py
```diff
@@ -0,0 +1,69 @@
...
+                "data-theme": "admin-autocomplete",
+                "data-allow-clear": json.dumps(not self.is_required),
+                "data-placeholder": "",  # Allows clearing of the input.
+                "lang": self.i18n_name,
+                "class": attrs["class"]
+                + (" " if attrs["class"] else "")
+                + "admin-autocomplete",
+            }
+        )
+        return attrs
```
patrick91, line 42: same as above

we could also move these to a base class if we are going to reuse them

backend/users/autocomplete.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def optgroups(self, name, value, attr=None):
+        default = (None, [], 0)
+        groups = [default]
+
+        if value and value[0]:
+            users_by_id = get_users_data_by_ids(value)
```
patrick91, line 49: do you think we can find a way to move this http call somewhere more visible?
I find it a bit worrying that it is hidden in this `optgroups` function

users-backend/users/internal_api/types.py
```diff
@@ -6,11 +6,18 @@
...
+    username: str
+    name: str
     email: str
     is_active: bool
     is_staff: bool
     jwt_auth_id: int
 
+    @strawberry.field
+    def display_name(self) -> str:
+        return self.fullname or self.name or self.username or self.email
```
patrick91, line 19: smart 😊

patrick91 (approved):

.github/workflows/pycon-backend-checks.yml
```diff
@@ -40,4 +40,4 @@ jobs:
         working-directory: backend
         run: poetry run python manage.py makemigrations --check
         env:
-          DJANGO_SETTINGS_MODULE: pycon.settings.dev
+          DJANGO_SETTINGS_MODULE: pycon.settings.test
```
estyxx, line 43: mmm why this change?

backend/submissions/admin.py
marcoacierno: I can see how it might look like

backend/users/admin_views.py
marcoacierno: I think if something goes wrong it should be fine to raise a 500 🤔

backend/users/autocomplete.py
marcoacierno: the default init wants the field as input

backend/users/autocomplete.py
marcoacierno: a few things are custom here, what is different is basically no reference to the field and the url is hardcoded there

you can see the original here: https://github.com/django/django/blob/8eb56930915f8022aa27ec29ab5be7d21764608c/django/contrib/admin/widgets.py#L396

backend/users/autocomplete.py
marcoacierno: `optgroups` is called the first time to prepare the UI to show what is already set

Not 100% sure what you find worrying about it behind inside optgroups? How would you make it more visibile?

.github/workflows/pycon-backend-checks.yml
marcoacierno: so that settings required in dev are not required during CI

backend/users/autocomplete.py
marcoacierno: see the base code here: https://github.com/django/django/blob/8eb56930915f8022aa27ec29ab5be7d21764608c/django/contrib/admin/widgets.py#L385

backend/users/autocomplete.py
marcoacierno: >we could also move these to a base class if we are going to reuse them

the base class AutocompleteSelect is not something we own but part of django, or are you referring to something else?

backend/users/autocomplete.py
marcoacierno: as far as I know we can't call this somewhere else, since we only have access to the initial form state here as far as I know

backend/users/autocomplete.py
patrick91: we can create our own base class :) but only if we use it more than once otherwise it is fine

backend/users/autocomplete.py
patrick91: I'd do the call outside and pass the data the autocomplete :) but it's not a big deal, as this is for the django admin

marcoacierno: uhm but we only have the data in here?

backend/users/autocomplete.py
marcoacierno: If we add more data from external services it might make sense yes, but ideally I would prefer the users service to be the only external service to call, otherwise we risk building a spider web of services calling each other :'D

backend/users/client.py
```diff
@@ -0,0 +1,29 @@
+from typing import Any
+
+from asgiref.sync import async_to_sync
+from django.conf import settings
+from pythonit_toolkit.service_client import ServiceClient
+
+GET_USERS_BY_IDS = """query GetUsersByIds($ids: [ID!]!) {
+    usersByIds(ids: $ids) {
+        id
+        displayName
```
estyxx, line 10: Can I use this query to get also the Gender of a user?

marcoacierno: yes I can add it

backend/submissions/admin.py
marcoacierno: done

PR #2478, opened 2021-11-18
WEB-134: Fix Voting rank

estyxx:


backend/voting/admin.py
```diff
@@ -110,14 +110,21 @@ def gender(self, obj):
...
+        speaker_gender = self._users_by_id[str(obj.speaker_id)]["gender"]
+        return emoji[speaker_gender]
 
     def view_submission(self, obj):  # pragma: no cover
         return format_html(
             '<a class="button" ' 'href="{{}}" target="_blank" >Open</a>&nbsp;',
             reverse("admin:submissions_submission_change", args=(obj.submission.id,)),
         )
 
+    def get_queryset(self, request):
```
marcoacierno, line 122: https://github.com/pythonitalia/pycon/pull/2477/files#

I implemented a mixin AdminUsersMixin that you can use so we don't need to copy&paste this get_queryset logic everywhere

backend/voting/admin.py
```diff
@@ -110,14 +110,21 @@ def gender(self, obj):
             "other": "🧑🏻‍🎤",
             "not_say": "⛔️",
         }
-        return emoji[obj.submission.speaker.gender]
+        speaker_gender = self._users_by_id[str(obj.speaker_id)]["gender"]
```
marcoacierno, line 113: maybe we can extend the mixin to have a simple `self.get_user_data(user_id)['speaker']`

marcoacierno: I merged it, you can do `get_user_data`

marcoacierno:
@estyxx can you rebase this PR?

backend/blog/admin.py
```diff
@@ -31,6 +31,6 @@ class PostAdmin(AdminUsersMixin):
     user_fk = "author_id"
 
     def author_display_name(self, obj):
-        return self.get_user_display_name(obj.author_id)
+        return self.get_user_data(obj.author_id)["displayName"]
```
marcoacierno, line 34: why are you changing this? 🤔

backend/users/mixins.py
```diff
@@ -12,8 +12,5 @@ def get_queryset(self, request):
         self._PREFETCHED_USERS_BY_ID = get_users_data_by_ids(list(users_ids))
         return queryset
 
-    def get_user_display_name(self, obj_id: Any) -> str:
-        return self.get_user_data(obj_id)["displayName"]
-
-    def get_user_data(self, obj_id: Any) -> str:
+    def get_user_data(self, obj_id: Any) -> dict[str, Any]:
```
marcoacierno, line 15: why remove get_user_display_name?

backend/blog/admin.py
estyxx: thought was better having one method to get the data... but maybe not?

marcoacierno: get_user_display_name is a utility method that calls get_user_data so we don't have to use ["displayName"] everywhere since that will be the most common use.

get_user_data is what you have to use since you have to access the gender field.

marcoacierno:
Your PR still contains my commits (weirdly?) did you rebase it?

estyxx:
I did a marge, as rebase was quite messy... I can reopen it maybe

marcoacierno:
uhm what happens if you do rebase interactive and just drop my commits?

marcoacierno:
any way it should be ok since we can squash merge

estyxx:
@marcoacierno I recreated the branch, do my changes and pushed them... now should be more clean :)

estyxx:
Do we want to test admins? or we should exclude it from the coverage? 🤔

marcoacierno:
it would be nice to test the admin too if it can be done since it is done and can be tested most of the time :) but in this case the tests missing are mostly for my mixin so don't worry about it, I will write them

backend/voting/admin.py
```diff
@@ -51,7 +51,7 @@ class Media:
 
 
 @admin.register(RankSubmission)
-class RankSubmissionAdmin(admin.ModelAdmin):
+class RankSubmissionAdmin(AdminUsersMixin):
```
marcoacierno, line 54: https://github.com/pythonitalia/pycon/blob/bced06dd103a41841904086cc5709e7aa70a36b4/backend/blog/admin.py#L31

you are missing the user_fk here, the mixin needs it to know where the user id is

```suggestion
class RankSubmissionAdmin(AdminUsersMixin):
    user_fk = 'speaker_id'
```

backend/voting/admin.py
```diff
@@ -51,7 +51,8 @@ class Media:
...
+    user_fk = "speaker_id"
```
marcoacierno, line 55: where is speaker_id coming from?

```
Cannot resolve keyword 'speaker_id' into field. Choices are: absolute_rank, absolute_score, id, rank_request, rank_request_id, submission, submission_id, topic_rank
```

we are getting this error now 🤔

marcoacierno: ok so RankSubmission has submission, so speaker_id doesn't exist but `submission__speaker_id` does

marcoacierno: I am not sure how it worked for you locally 🤔

estyxx: I did a little change as the `speaker_id` is in the `submission`... can you try again? tnks

backend/voting/admin.py
```diff
@@ -107,7 +107,7 @@ def gender(self, obj):
             "not_say": "⛔️",
         }
 
-        speaker_gender = self.get_user_data(obj.speaker_id)["gender"]
+        speaker_gender = self.get_user_data(obj.submission.speaker_id)["gender"]
```
marcoacierno, line 110: You also have to change user_fk

backend/voting/admin.py
marcoacierno: you have to change user_fk to `submission__speaker_id` otherwise it will just fail when trying to fetch the list of users

backend/voting/admin.py
```diff
@@ -51,7 +51,8 @@ class Media:
 
 
 @admin.register(RankSubmission)
-class RankSubmissionAdmin(admin.ModelAdmin):
+class RankSubmissionAdmin(AdminUsersMixin):
+    user_fk = "submission.speaker_id"
```
marcoacierno, line 55:
```suggestion
    user_fk = "submission__speaker_id"
```

this fk goes to a query, doesn't access the object

marcoacierno (approved):
image

Happy to merge this!

PR #2479, opened 2021-11-18
Show again buy tickets CTA

marcoacierno:
This PR shouldn't be merged before the 1st of December, when tickets go live.


marcoacierno:
image

we need to change the logo :P

frontend/src/pages/[lang]/index.tsx
```diff
@@ -43,7 +43,7 @@ export const HomePage = () => {
       <FormattedMessage id="home.title">
         {(text) => <MetaTags title={text} />}
       </FormattedMessage>
-      <HomepageHero hideBuyTickets={true} />
+      <HomepageHero hideBuyTickets={false} />
```
Comment (estyxx) [inline frontend/src/pages/[lang]/index.tsx:46]: Can we add an env variable a flag maybe later on? Would be easier to hide/show this kind of stuff??

Comment (marcoacierno) [reply frontend/src/pages/[lang]/index.tsx:46]: we can add a deadline of type tickets! that would be better

Comment (estyxx) [reply frontend/src/pages/[lang]/index.tsx:47]: Uh yeah! we should add a ticket for this !

PR #2480, opened 2021-11-19
Fix marquee on the homepage not looping correctly

marcoacierno:
## Why

currently our marquee doesn't loop correctly and in some screen sizes it doesn't cover all the text

## How to test it

Go to https://pycon-git-fix-marquee-python-italia.vercel.app/ and look at it looping

### DIFF
--- a/frontend/src/components/marquee/index.tsx
+++ b/frontend/src/components/marquee/index.tsx
@@ -18,6 +18,13 @@ const animation = keyframes`
   }
 `;
 
+// This is the font width of the character 0 at the font size of 32px
+// It should be roughly the same as a CSS 1ch value.
+// To calculate it I used this solution: https://stackoverflow.com/a/21015393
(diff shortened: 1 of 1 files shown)


patrick91 (approved):

frontend/src/components/marquee/index.tsx
```diff
@@ -2,40 +2,82 @@
...
       sx={{
-        willChange: "transform",
-        animation: `${animation} 20s linear infinite`,
+        height: 125,
+        alignItems: "center",
+        borderTop: "primary",
+        borderBottom: "primary",
       }}
-      css={`
-        @media (prefers-reduced-motion: reduce) {
```
marcoacierno, line 33: I need to add this back

marcoacierno: done

PR #2499, opened 2021-11-26
Remove notes from submission query

estyxx:


patrick91:
I think we should change the backend to return an empty string (or `None`) if they are not allowed to see the notes

backend/api/submissions/types.py
```diff
@@ -73,6 +72,12 @@ class Submission:
         LazyType["AudienceLevel", "api.conferences.types"]
     ] = restricted_field()
 
+    @strawberry.field
+    def notes(self, info) -> Optional[str]:
+        if CanSeeSubmissionPrivateFields().has_permission(self, info):
+            return self.notes
+        return None
```
marcoacierno, line 79: should we add tests for this?

backend/api/submissions/types.py
```diff
@@ -73,6 +72,12 @@ class Submission:
         LazyType["AudienceLevel", "api.conferences.types"]
     ] = restricted_field()
 
+    @strawberry.field
+    def notes(self, info) -> Optional[str]:
+        if CanSeeSubmissionPrivateFields().has_permission(self, info):
+            return self.notes
```
marcoacierno, line 78: I am thinking... do we want this behaviour for all `private_field`? maybe we could implement a generic solution inside `private_field` and `restricted_field`?

estyxx: Yeah probably, maybe in the future we want to show all the details on the Submission page if the user is an admin. Maybe when we are preparing the schedule...

estyxx: I was thinking to do something like this: 

```python
def private_field(field_name):
    """Field that can only be seen by admin and the submitter"""

    def resolver(self, info):
        if CanSeeSubmissionPrivateFields().has_permission(self, info):
            return getattr(self, field_name)
        return None

    return strawberry.field(resolver=resolver)


@strawberry.type
class Submission:
	...
    notes: Optional[str] = private_field("notes")
```

But I was thinking if there is a clever way to get the name of the field... and don't have to pass to the `private_filed` method... I'm pretty sure internally strawberry has to do something to retrieve it somehow...

marcoacierno:
Will have a look tomorrow!

Also I think you need to update the tests

backend/api/submissions/types.py
```diff
@@ -59,19 +71,21 @@ class Submission:
...
     title: str
     slug: str
-    elevator_pitch: Optional[str] = restricted_field()
-    abstract: Optional[str] = restricted_field()
-    speaker_level: Optional[str] = private_field()
-    previous_talk_video: Optional[str] = private_field()
-    notes: Optional[str] = private_field()
-    topic: Optional[LazyType["Topic", "api.conferences.types"]] = restricted_field()
-    type: Optional[SubmissionType] = restricted_field()
+    elevator_pitch: Optional[str] = restricted_field("elevator_pitch")
```
marcoacierno, line 74: Uhm do we maybe have access to the field name into `info`?

estyxx: but how do I get `info` in my decorator?

patrick91: you have it here: https://github.com/pythonitalia/pycon/pull/2499/files#diff-aacfacd937601019c5760248c509135b31e5b433a0aa89694434c5e8b634b8e1R18 😊

estyxx: mmm I tried with `info.field_name` but seems to be in camelCase... 🤔 
```
AttributeError: 'Submission' object has no attribute 'elevatorPitch'
```

estyxx: Ok i used `python_name` (somehow locally I can't see it as a property of Info object... )

backend/api/submissions/types.py
```diff
@@ -11,15 +13,27 @@
...
-def private_field():
+    return strawberry.field(resolver=resolver)
+
+
+def private_field() -> StrawberryField:
     """Field that can only be seen by admin and the submitter"""
-    return strawberry.field(permission_classes=[CanSeeSubmissionPrivateFields])
+
+    def resolver(self, info: Info):
+        if CanSeeSubmissionPrivateFields().has_permission(self, info):
```
marcoacierno, line 32: Can you fix https://github.com/pythonitalia/pycon/blob/5e3d8be0418ac70213112e000c306a3100ef3434/backend/api/submissions/permissions.py#L45 in this PR too?

So private_field are supposed to be seen only by admin and submitter, but right now there is:

```python
        if source.schedule_items.exists():  # pragma: no cover
            return True
```

which means that once it is added to the schedule you can fetch the private notes.

You only need to remove that `if`, if you can do it in this PR it would be great :)

marcoacierno: actually we might need to move that check into `CanSeeSubmissionDetail`

marcoacierno: @estyxx can you do this too? ^

marcoacierno (approved):

PR #2507, opened 2021-11-26
Add initial API for tickets left

patrick91:
This PR adds an API to show the quantity left for tickets :)

We need to update pretix for this to work :)

### DIFF
--- a/backend/api/pretix/query.py
+++ b/backend/api/pretix/query.py
@@ -39,6 +39,9 @@ def _get_category_for_ticket(item, categories):
 
 
 def _get_quantity_left_for_ticket(item, quotas):
+    if item["show_quota_left"] is False:
+        return None
+
     return sum(
         quota["available_number"]
         for quota in quotas.values()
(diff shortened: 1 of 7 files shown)


backend/api/pretix/query.py
```diff
@@ -38,7 +38,18 @@ def _get_category_for_ticket(item, categories):
     return categories.get(category_id)
 
 
-def _create_ticket_type_from_api(item, id, categories, questions, language):
+def _get_quantity_left_for_ticket(item, quotas):
+    if item["show_quota_left"] is False:
+        return None
+
+    return sum(
```
marcoacierno, line 45: I don't think `sum` is correct here, I think this should be `min`, because if you have ticket in 2 quotas (A and B), and quota A has 10 available and quota B only has 1 available, you can only buy it once since otherwise you would oversell quota B

patrick91: interesting! ok, will update :)

patrick91:
```suggestion
    # tickets can be in multiple quotas, in that case the one that has the least amount of tickets
    # should become the source of truth for availability. See:
    # https://docs.pretix.eu/en/latest/development/concepts.html#quotas
    return min(
```

marcoacierno (approved):

PR #2519, opened 2021-11-30
Try fixing CI build of Docker images

marcoacierno:
## Why

CI is currently failing because some file do not exist. I assume the native dependencies changed name

## How to test it

/deploy
it should build


patrick91 (approved):
+1

PR #2525, opened 2021-12-04
Migration to Next.js locale system

marcoacierno:
## Why

Native intl URL support, instead of using params

## How to test it

Vercel URL


backend/api/helpers/maps.py
```diff
@@ -21,6 +22,7 @@ def generate_map_image(
 
 @strawberry.type
 class Map:
+    id: ID
```
marcoacierno, line 25: added IDs so Apollo can re-use cache better when switching between it and en

frontend/src/components/header/header.graphql
```diff
@@ -1,23 +1,45 @@
...
-    slug(language: $language)
-  }
-}
-
-query Header($code: String!, $language: String!) {
+query Header($code: String!) {
   conference(code: $code) {
     id
-    conferenceMenu: menu(identifier: "conference-nav") {
+    conferenceMenuEn: menu(identifier: "conference-nav") {
```
marcoacierno, line 4: this allows Apollo to reuse the query result and not have to fetch the menus every time we switch language

frontend/src/components/header/index.tsx
```diff
@@ -27,32 +27,28 @@ const ProfileLink = dynamic(
...
   links: { href: string; title: string; page?: { slug: string } | null }[];
-}> = ({ links }) => (
+}) => (
   <Fragment>
     {links.map((link) => {
-      let path = link.page ? "/[lang]/[slug]" : link.href;
-
-      // nasty hack to make client side routing work with next.js
-      path = path.replace("/en", "/[lang]").replace("/it", "/[lang]");
+      const path = link.href;
```
marcoacierno, line 51: Will add (or remove?) page support better in another PR. I am writing it down and then write a ticket

frontend/src/components/link/index.tsx
```diff
@@ -81,12 +82,13 @@ export const Link: React.FC<LinkProps> = ({
...
   external = false,
   params = null,
+  locale,
   ...additionalProps
 }) => {
   const language = useCurrentLanguage();
 
   if (!url) {
-    url = path.replace("[lang]", language);
+    url = path;
```
marcoacierno, line 91: I am writing this down too. I think the link component is a bit messy right now with lots of href/target/variant/path that are not documented and hard to understand

frontend/src/components/schedule-view/events.tsx
```diff
@@ -106,12 +106,12 @@ export const Submission = ({
...
   if (item.type === "submission") {
-    return `/[lang]/talk/[slug]`;
+    return `/talk/[slug]`;
   }
 
   // TODO: check tbd
   if (item.type === "keynote") {
-    return `/[lang]/keynote/[slug]`;
+    return `/talk/[slug]`;
   }
```
marcoacierno, line 115: the keynote page doesn't exist, but it is a good idea. For now I added a link to the talk page

writing this down because this also implies that the keynote will be a submission in BE

frontend/src/helpers/prefetch.ts
```diff
@@ -4,11 +4,10 @@ import { queryFooter, queryHeader } from "~/types";
 
 export const prefetchSharedQueries = async (
   client: ApolloClient<any>,
-  language: string,
+  _language: string,
```
marcoacierno, line 7: language is not needed anymore since we fetch all menu items now. but I don't want to remove it since it might be better in future and don't want to edit 100 files again 😓

frontend/src/locale/context.tsx
```diff
@@ -1,46 +1,28 @@
...
 import cookies from "next-cookies";
+import { useRouter } from "next/router";
 
 import { Language } from "~/locale/languages";
 
-interface ContextProps {
-  readonly locale: string;
-}
-
-export const AlternateLinksContext = React.createContext({
```
marcoacierno, line 11: all of this is no longer needed. Next.js is able now to handle this

frontend/src/locale/context.tsx
```diff
@@ -1,46 +1,28 @@
...
+}: React.PropsWithChildren<{ lang: string }>) => {
+  const language = useCurrentLanguage();
+
   React.useEffect(() => {
     const { pyconLocale } = cookies({
       req: { headers: { cookie: document.cookie } },
     });
-    if (lang !== pyconLocale) {
-      document.cookie = `pyconLocale=${lang}; path=/`;
+    if (language !== pyconLocale) {
```
marcoacierno, line 17: this component now is only responsible to update the local user cookie

frontend/src/pages/schedule/index.tsx
```diff
@@ -0,0 +1,18 @@
+import Head from "next/head";
+import { useRouter } from "next/router";
+
+export const SchedulePage = () => {
+  const { replace } = useRouter();
+
+  // TODO Replace with _middleware
+  if (typeof window !== "undefined") {
+    replace("/schedule/2022-06-02");
+  }
```
marcoacierno, line 10: as the comment said, I will create a ticket, but there is a bug (?) with nextjs that broke `redirect: {}` so now we can't use it anymore

I want to use _middleware but Apollo raises an error due to undefined __DEV__, which seems to be removed due to tree-shaking. but even updating apollo didn't fix it :(

frontend/src/pages/_middleware.ts
```diff
@@ -3,15 +3,20 @@ import { NextResponse } from "next/server";
...
   const locale = getLocale(req.cookies.pyconLocale);
 
-  if (pathname === "/") {
-    return NextResponse.redirect(`/${locale}`);
-  }
-
-  return NextResponse.next();
+  return shouldHandleLocale
+    ? NextResponse.redirect(
+        `/${locale}${req.nextUrl.pathname.replace("/default", "")}`,
```
simobasso, line 17: What is the purpose of this? We really need `default` as a default locale?

frontend/src/pages/index.graphql
```diff
@@ -21,20 +21,25 @@ query IndexPage($code: String!, $language: String!) {
     gettingThereText: copy(key: "getting-there-text", language: $language)
 
     map {
+      id
+      image(width: 1280, height: 400, zoom: 15)
```
simobasso, line 25: Ok the `id` but what about this `image`? Is intended?

simobasso (commented):
Nice work! 👏

frontend/src/components/day-selector/index.tsx
```diff
@@ -32,10 +31,7 @@ export const DaySelector: React.FC<{
         <Select
           sx={{ mt: 3, width: "100%" }}
           onChange={(e: React.ChangeEvent<HTMLSelectElement>) =>
-            router.push(
-              "/[lang]/schedule/[day]",
-              getDayUrl(language, e.target.value),
-            )
+            router.push("/schedule/[day]", getDayUrl(language, e.target.value))
```
estyxx, line 34: Do we still need language here??

frontend/src/pages/_middleware.ts
estyxx: think this should be the workaround to make Next.js include the `/en/` prefix in the default locale
https://nextjs.org/docs/advanced-features/i18n-routing#prefixing-the-default-locale

frontend/next.config.js
```diff
@@ -18,6 +18,12 @@ const {
 } = process.env;
 
 module.exports = withSourceMaps({
+  i18n: {
+    locales: ["default", "en", "it"],
+    defaultLocale: "default",
+    localeDetection: false,
```
patrick91, line 24: just curious, why is this false?

frontend/src/components/cfp-send-submission/index.tsx
```diff
@@ -74,7 +72,7 @@ export const CfpSendSubmission: React.SFC = () => {
 
     if (response.data?.mutationOp.__typename === "Submission") {
       const id = response.data.mutationOp.id;
-      router.push(`/[lang]/submission/[id]`, `/${lang}/submission/${id}`);
+      router.push(`/submission/[id]`, `/submission/${id}`);
```
patrick91, line 75: that's cool!

frontend/src/components/day-selector/index.tsx
patrick91: I guess marco left it so he didn't have to change all the places using getDayUrl :D

frontend/src/components/header/header.graphql
```diff
@@ -1,23 +1,45 @@
...
+        }
+      }
+    }
+    programMenuIt: menu(identifier: "program-nav") {
+      links {
+        title(language: "it")
+        href(language: "it")
+
+        page {
+          slug(language: "it")
```
patrick91, line 41: do you think in future we should move the language to be a header? We can chat about this on slack, just curious of what you think :D

frontend/src/pages/[lang]/schedule/index.tsx
```diff
@@ -1,39 +0,0 @@
-import { GetStaticPaths, GetStaticProps } from "next";
```
Comment (patrick91) [inline frontend/src/pages/[lang]/schedule/index.tsx:1]: why was this deleted?

frontend/src/pages/_middleware.ts
```diff
@@ -3,15 +3,20 @@ import { NextResponse } from "next/server";
 
 import { DEFAULT_LOCALE, VALID_LOCALES } from "~/locale/languages";
 
+const PUBLIC_FILE = /\.(.*)$/;
+
 export function middleware(req: NextRequest, _ev: NextFetchEvent) {
-  const { pathname } = req.nextUrl;
+  const shouldHandleLocale =
+    !PUBLIC_FILE.test(req.nextUrl.pathname) &&
+    !req.nextUrl.pathname.includes("/api/") &&
```
patrick91, line 11: I think we also have `/admin` (not sure if is set on next js as a redirect though)

frontend/src/pages/_middleware.ts
```diff
@@ -3,15 +3,20 @@ import { NextResponse } from "next/server";
...
-  if (pathname === "/") {
-    return NextResponse.redirect(`/${locale}`);
-  }
-
-  return NextResponse.next();
+  return shouldHandleLocale
+    ? NextResponse.redirect(
+        `/${locale}${req.nextUrl.pathname.replace("/default", "")}`,
+      )
+    : undefined;
```
patrick91, line 19: very nit: I'd use an if instead of a ternary operator here, especially because we are returning undefined in one case :)

frontend/src/pages/[lang]/schedule/index.tsx
Comment (patrick91) [reply frontend/src/pages/[lang]/schedule/index.tsx:1]: just found it below :)

patrick91 (approved):

frontend/src/components/day-selector/index.tsx
marcoacierno: 🤔 let me check

frontend/src/components/header/header.graphql
marcoacierno: having the language in the header means that apollo won't be able to cache based on the language 🤔

patrick91: yup, but maybe there's a way to prevent that, that's something we could try in future, mostly because I'm curios to see if it is possible 😊

frontend/src/pages/index.graphql
marcoacierno: Yes it is intended because I wanted to fetch everything together, but since we now have an ID I think I can remove it and have the MapWithLink run the query and thanks to the ID Apollo should be able to merge the two

frontend/src/pages/_middleware.ts
marcoacierno: The reason there is the pathname.replace is because of an issue with how Nextjs returns .href: https://github.com/vercel/next.js/issues/31533

frontend/next.config.js
marcoacierno: it was suggested in the nextjs section: https://nextjs.org/docs/advanced-features/i18n-routing#prefixing-the-default-locale we can try enabling it and see if it has any issue. I have a few ideas on why it might be required but I need to investigate before saying anything

frontend/src/components/header/header.graphql
marcoacierno: This is an issue I found a few days ago of someone that has the language in the header: https://github.com/apollographql/apollo-client/issues/1420 and ended up having to send the language as variable to fix the cache

frontend/src/pages/_middleware.ts
marcoacierno: 👍

frontend/src/pages/_middleware.ts
marcoacierno: it is set in now.json but it seems it gets ignored now 🤔

patrick91: I wouldn't bother with it :D

marcoacierno: I fixed it, for some weird reason now.json (or vercel.json) were ignored, I moved it into next.config.js

estyxx (commented):
Good work! 💪🏻

estyxx (approved):

PR #2527, opened 2021-12-05
web-172 - Show CFP deadline in the introduction

estyxx:

### DIFF
--- a/frontend/src/pages/cfp/index.tsx
+++ b/frontend/src/pages/cfp/index.tsx
@@ -1,7 +1,7 @@
 /** @jsxRuntime classic */
 
 /** @jsx jsx */
-import { Fragment } from "react";
+import { Fragment, useMemo } from "react";
 import { FormattedMessage } from "react-intl";
 import { Box, Container, Heading, jsx, Text } from "theme-ui";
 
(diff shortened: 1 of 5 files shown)


frontend/src/pages/[lang]/cfp/is-cfp-open.graphql
```diff
@@ -2,5 +2,9 @@ query IsCfpOpen($conference: String!) {
   conference(code: $conference) {
     id
     isCFPOpen
+    deadlines {
```
Comment (marcoacierno) [inline frontend/src/pages/[lang]/cfp/is-cfp-open.graphql:5]: you can fetch a single deadline: https://github.com/pythonitalia/pycon/blob/5d0cac9834203f96204e423eb0b797c9704c041a/frontend/src/pages/index.graphql#L29

Comment (estyxx) [reply frontend/src/pages/[lang]/cfp/is-cfp-open.graphql]: Oh yeah, of course, I didn't see it!

frontend/src/pages/cfp/index.tsx
```diff
@@ -57,18 +58,24 @@ const CfpSectionOrClosedMessage: React.SFC<{ open: boolean }> = ({ open }) => {
...
-
   return (
     <Fragment>
       <FormattedMessage id="cfp.pageTitle">
         {(text) => <MetaTags title={text} />}
       </FormattedMessage>
-      <Introduction />
+      <Introduction
+        deadline={formatDeadlineDate(
+          data?.conference.cfpDeadline?.end,
```
marcoacierno, line 75: this doesn't seem correct, if we pass undefined `new Date()` would fail 🤔

marcoacierno: image
this is what happens 🤔  we need to improve this CFP page after we merge my PR #2540

PR #2540, opened 2021-12-10
Make grants form more dynamic via a deadline

marcoacierno:
## Why

Grants page should be smarter so we can share it even before we open the form

## How to test it

Clone the PR
Go to http://localhost:8000/admin/conferences/conference/1/change/
Create a new type of deadline "Grants"
Go to http://localhost:3000/grants
You will see different messages based on the date you used

Grant form is closed:

image

Grant form will open in the future:

image

Grant form open:

image

If the deadline is not set we return a 404

image


patrick91 (approved):
looks good!

PR #2542, opened 2021-12-11
Remove default options

estyxx:


frontend/src/components/cfp-form/index.tsx
```diff
@@ -91,41 +109,6 @@ export const CfpForm: React.SFC<Props> = ({
       },
     );
 
-  const setupCleanForm = (data: CfpFormQuery) => {
```
marcoacierno, line 94: you can remove the `CfpFormQuery` import now (ther is an eslint warning)

frontend/src/components/cfp-form/index.tsx
```diff
@@ -152,13 +135,8 @@ export const CfpForm: React.SFC<Props> = ({
           "tags",
           submission!.tags.map((t) => t.id),
         );
-        formState.setField(
-          "speakerLevel",
-          submission!.speakerLevel || SPEAKER_LEVEL_NEW_VALUE,
-        );
+        formState.setField("speakerLevel", submission!.speakerLevel);
```
marcoacierno, line 138: uhm we used to have the `||` because `speakerLevel` was a field added later, but I think now it is not needed anymore 👍

marcoacierno (approved):
Looks good. We have an issue with use-form not working when editing fields that we can fix in another PR

PR #2544, opened 2021-12-11
Move submission edit load outside Apollo flow

marcoacierno:
## Why

Submission edit is not working anymore

## How to test it

Send a CFP, try editing it. It should show the form with all the values in it

### DIFF
--- a/frontend/src/components/cfp-form/index.tsx
+++ b/frontend/src/components/cfp-form/index.tsx
@@ -116,28 +116,6 @@ export const CfpForm: React.SFC<Props> = ({
     variables: {
       conference: conferenceCode,
     },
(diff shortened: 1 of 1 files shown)


frontend/src/components/cfp-form/index.tsx
```diff
@@ -116,28 +116,6 @@ export const CfpForm: React.SFC<Props> = ({
     variables: {
       conference: conferenceCode,
     },
-    onCompleted(data) {
```
marcoacierno, line 119: not 100% sure why this stopped working.

The same code works in the profile edit 🤷 

My hypothesis is that Apollo onCompleted flow changed somehow

PR #2547, opened 2021-12-11
Add support to become member of Python Italia on pycon.it

marcoacierno:
## Why

Step #3 of allowing people to subscribe to python italia via pycon.it
This is the last required step we need before releasing this feature

A few things aren't ideal here:

1. People should not be able to subscribe if they are already members. This check is only done via FE right now. Ideally this should be blocked on the BE too.
2. You can only add the membership to the order once. Again, this is done by FE. Should be checked on the BE too.

The two points above aren't ideal but I fear the code to handle those would be very ugly, so I think I would prefer rewriting the whole ticketing system to make it more generic instead of hacking those checks in 🤔 

## How to test it

Go to http://localhost:3000/en/tickets

User is already a member:

image

user can buy the membership

image


backend/api/pretix/query.py
```diff
@@ -146,9 +148,11 @@ def get_conference_tickets(
     quotas = pretix.get_quotas(conference)
 
     def sort_func(ticket):
-        # If the item has variations, it means it is the
-        # t-shirt product. We want to show it always at the bottom
-        if len(ticket.variations) > 0:
+        # Make gadgets and association appear at the end
```
estyxx, line 151: Maybe we can turn the env variable into a list containing both `["Gadget", "Association"]` so if we add more categories in the future at the bottom we don't have to change the code?

frontend/src/components/hotel-form/index.tsx
```diff
@@ -23,6 +23,7 @@ type Props = {
     checkout: moment.Moment,
   ) => void;
   removeHotelRoom: (id: string, index: number) => void;
+  me: CurrentUserQueryResult["data"]["me"];
```
estyxx, line 26: Why you can't use the dot notation here? `CurrentUserQueryResult` is a type, isn't it? 🤔

frontend/src/components/tickets-form/product-row.tsx
```diff
@@ -170,6 +175,15 @@ export const ProductRow = ({
           </div>
         )}
 
+        {!ticket.soldOut && isMembershipProduct && (
```
estyxx, line 178: This `ticket` is the one for the association subscription? Can be it SoldOut? 
If the tickets are sold out, people can't subscribe from this page to the association?

marcoacierno: Nope, it can't go sold-out, the soldOut check is there only because I copied and pasted the code.

The whole tickets page is currently a huge mess so I want to refactor it in the future

backend/api/pretix/query.py
marcoacierno: it isn't an environment variable but a constant defined here: https://github.com/pythonitalia/pycon/pull/2547/files#diff-5762a1b4022b3f800641216902ef7c088ac8a70c0d37057eff136650db2cb697R1

marcoacierno: I don't have a preference to be honest.

If I were to change this code I would split the if instead of using `or` to make it more clear

frontend/src/components/hotel-form/index.tsx
marcoacierno: Uhm can we? This is the typescript definition, not code, basically we are telling TS here that the type of `me` is the type of `CurrentUserQueryResult > data > me`

image

marcoacierno: https://www.typescriptlang.org/play?ssl=9&ssc=1&pln=9&pc=2#code/FAFwngDgpgBAgjAvDA3sGMCGAuGBnEAJwEsA7AcwG5gBfYUSWAISVXRhCgNzgG0AiTPwC61DJwIAmHgDpMtIA

Doesn't seem to be supported

frontend/src/components/tickets-page/wrapper.tsx
```diff
@@ -54,7 +55,36 @@ export const TicketsPageWrapper: React.SFC<Props> = ({ children }) => {
...
+          continue;
+        }
+
+        removeProduct(product.id);
+        ticketsHaveBeenUpdated = true;
+      }
+
+      if (ticketsHaveBeenUpdated && typeof window !== "undefined") {
+        // This is an hack because the products are not correctly removed :(
+        window.location.reload();
```
marcoacierno, line 84: this is terrible

PR #2549, opened 2021-12-12
Refactor Submission's permissions

estyxx:
## Why

Was a mess, hope now the tests are more clear :)


backend/api/submissions/permissions.py
```diff
@@ -27,12 +28,10 @@ def has_permission(self, source, info):
         if user_info.is_staff or source.speaker_id == user_info.id:
             return True
 
-        if Submission.objects.filter(
-            speaker_id=user_info.id, conference=conference
-        ).exists():
-            return True
+        if conference.is_voting_closed:
```
estyxx, line 31: Moved this here below because we don't want the fields to be visible if the talk was not accepted, and if the user could have seen the talk (because, admin or author) should be already fall into one of the conditions above

backend/voting/tests/test_helpers.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+@pytest.mark.xfail
+def test_user_can_vote_if_has_bought_a_ticket_in_past_editions():
+    assert False
+
+
+@pytest.mark.xfail
+def test_user_can_vote_if_is_a_member_of_python_italy():
+    assert False
```
marcoacierno, line 43: if we want to merge this PR, I would remove those tests and add them back when we implement the two cases

backend/api/submissions/permissions.py
```diff
@@ -27,12 +28,10 @@ def has_permission(self, source, info):
         if user_info.is_staff or source.speaker_id == user_info.id:
             return True
```
marcoacierno, line 29: can you move all those if(s) check that don't do queries above the schedule_items?

I am talking about:

```
if not pastaporto.is_authenticated:
```

and

```
        if user_info.is_staff or source.speaker_id == user_info.id:
```

seems wasteful to run the schedule_items query if we are then going to pass the Permission because they are staff or reject because they are unauthenticated

estyxx: I don't think that is right, because `if source.schedule_items.exists()` means the submission is public, and so anyone can see it because it's in the Schedule, even if they don't have an account and are not authenticated. 
If I swap those checks, I will force them to have an account... doesn't seem right to me...

estyxx: Any thoughts on this?

marcoacierno: >  If I swap those checks, I will force them to have an account... doesn't seem right to me...

Ideally what I wanted was that the code was smart enough that if you are logged in > do the easy check that don't require the DB first

If you are not logged in and it is not public, fail 😃 


But don't worry, it is a minor optimization and we can do it in another PR

marcoacierno (approved):

PR #2553, opened 2021-12-14
Rewrite association membership structure and domain

marcoacierno:
## Test

Vercel URL: https://association-frontend-qyv04d9pp-python-italia.vercel.app/

Create a staging account, subscribe to the association (stripe test cards), everything should work and look like before

## Explanation

This refactor changes how we store membership on our DB. The original MVP was 1:1 with Stripe meaning that it wasn't flexible enough for what we want to do now, which is allowing our customers to join our association for a year by buying it on Pretix when buying their pycon.it tickets.

The new structure changes the service to be around the concept of "Subscription". There can only be 1 subscription per user in our system. A user is either a member or not.
If there isn't a subscription, or the subscription status is CANCELED, the user is not a member.
We want to allow a user to be subscribed to the association via different methods: either via the stripe subscription system or pretix payments.

To allow it, the service keeps track of two major entities in it is domain: **Subscription** and **Payment**.

A **Subscription** is just an entity connected to a user that says if a person is either a member or not. (Via a status field)

A **Payment** is an entity that keeps track of all payments associated with the subscription. Payments are what allow a subscription to become **ACTIVE**. A Payment has a **period** which allows us to understand what period of membership that payment is supposed to cover. Right now all periods are of 1 year, in the future we could allow shorter or longer periods.

A **Payment** in our domain exists in two ways (for now): a **PretixPayment** and a **StripeSubscriptionPayment**.

A **PretixPayment** is a payment coming from pretix from an event in there, we keep additional info like the order_code to know from which order this membership is coming from.

A **StripeSubscriptionPayment** is a payment coming from the Stripe Subscription system, it keeps additional information like the invoice_id, the pdf and the subscription it originated from. A user can freely cancel their stripe subscription, create a new one after 1 year.

All payments keep track of how much the user has paid and a **idempotency key** which is used to make sure we don't process the same order twice. For example in Stripe it is the invoice ID, for pretix could be a combination of "event_organiser+event_slug+order_code". This allow us to receive the same event multiple times and not count the same payment twice.

To handle stripe payments, our domain also has the concept of **StripeCustomer**, used to link the user and their equivalent on Stripe. Ideally our users only exist once on our Stripe account.

When the user is subscribed via Stripe, they can manage their subscription too (like changing their payment method or canceling it), this is supposed via the **StripeCustomer** entity.

Info about the approach: https://www.figma.com/file/xQ5ewPKlkd8zW3SpShjH56/Association---Subscription-from-pycon.it-tickets?node-id=38%3A285

---

Generic TODOs (for other PRs)

- [ ] Cronjob checking for expired subscriptions (new PR?)
- [x] Fix tests
- [ ] Update the FE to show the manage feature only if the user uses stripe
- [x] Deploy on staging, connect to DB and create a new DB
- [ ] On production manual data migration creating the StripeCustomer and Payments for existing members


association-backend/src/association_membership/domain/entities.py
```diff
@@ -62,40 +68,81 @@ def _change_state(self, to: SubscriptionStatus):
         logger.info("Switching subscription from status %s to %s", self.status, to)
         self.status = to
 
+    def add_stripe_subscription_payment(
```
estyxx, line 71: What about `add_pretix_payment`, is for another PR?

patrick91 (approved):

association-backend/src/association_membership/domain/entities.py
```diff
@@ -30,23 +42,17 @@ class Meta(BaseMeta):
...
 
-    _add_invoice: List[SubscriptionInvoice] = PrivateAttr(default_factory=list)
-
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        super().__init__(*args, **kwargs)
-        self._add_invoice = []
-
-    def add_invoice(self, payment: SubscriptionInvoice):
-        self._add_invoice.append(payment)
+    _add_stripe_subscription_payment: List[StripeSubscriptionPayment] = PrivateAttr(
```
estyxx, line 53: What's the purpose of this private attribute? it's the Orm that wants it? Seems unnecessary otherwise...

marcoacierno: see here: https://github.com/pythonitalia/pycon/pull/2553/files/6fea3cf05683082eac8f6fe54300b3074a707da5#diff-700c143ae661c70592ba9db5004cf76292fc6b55efbf54ff109adcc1be15f846R97

this is used to store what objects should be added when saving the object in the repository.

This allows us to separate DB operations from Domain operations

association-backend/src/association_membership/domain/entities.py
marcoacierno: Yes, I added the model already (which I might regret) but don't want to add more since we don't really know how the pretix integration looks like

estyxx (approved):

PR #2557, opened 2021-12-17
Add support to buying an Association Membership via Pretix

marcoacierno:
## Why

We want to allow people to subscribe to the association on pycon.it
This is part 2 of implementing this, this is a webbook that takes an order paid event from Pretix, checks if there is a position where the item is in a category with the internal name "Association", if there is, we take the paid payment and create a subscription from the payment date.

Pretix doesn't give us the user id, so we need to take the email from the order and find the user id

The rest is pretty similar, we create a Pretix payment from that as PAID, check that the payment date is within the current date (this is for safety in case we receive old events) and if it is, mark the subscription as active

In the future we might want to handle `pretix.event.order.canceled` to update the payment state if it gets canceled/refunded/anything else

## How to test it

Send a POST request to http://localhost:8060/pretix-webhook with 

payload:

```json
{"notification_id": 4122, "organizer": "test-organizer", "event": "local-conf-test", "code": "9YKZK", "action": "pretix.event.order.paid"}
```

auth:

username: `pretix`
password: `secret`

to test with pretix, configure a webhook here: https://tickets.pycon.it/control/organizer/test-organizer/webhooks
use ngrok to connect your local BE

the URL might look like this:

```
 http://pretix:secret@906b-2a01-4f8-252-ea-00-2.ngrok.io/pretix-webhook 
```

select `pretix.event.order.paid` as event.


association-backend/src/webhooks/handlers/pretix/api.py
```diff
@@ -0,0 +1,38 @@
+from typing import Literal
+
+import httpx
+
+from src.association.settings import PRETIX_API_TOKEN, PRETIX_API_URL
+
+METHODS = Literal["get"]
+
+
+class PretixAPI:
```
marcoacierno, line 10: this will soon live in toolkit and be reused between pycon backend and association: https://linear.app/python-italia/issue/WEB-211/muovere-linterfaccia-api-pretix-in-toolkit

association-backend/src/webhooks/handlers/pretix/pretix_event_order_paid.py
```diff
@@ -0,0 +1,161 @@
...
+        result["id"]: result for result in pretix_api.get_categories()["results"]
+    }
+
+    order_positions = order_data["positions"]
+    membership_positions = []
+    for position in order_positions:
+        item_data = pretix_api.get_item_data(position["item"])
+        category = categories.get(item_data["category"], None)
+
+        assert (
```
estyxx, line 44: We are expecting to receive an error? For the other items of the order that are for the event and not for the association, I guess... ? 
How this is supposed to work?

association-backend/src/webhooks/handlers/pretix/pretix_event_order_paid.py
```diff
@@ -0,0 +1,161 @@
...
+            "the payment with key=%s from order_code=%s",
+            action,
+            organizer,
+            event,
+            idempotency_key,
+            order_code,
+        )
+        return
+
+    subscription = await repository.get_user_subscription(user_id)
```
estyxx, line 99: What about the case when the user already has a subscription but it's not active?

association-backend/src/webhooks/tests/handlers/pretix/payloads.py
```diff
@@ -0,0 +1,347 @@
...
+        "vat_id": "",
+        "vat_id_validated": False,
+        "internal_reference": "",
+    },
+    "positions": [
+        {
+            "id": 801,
+            "order": "9YKZK",
+            "positionid": 1,
+            "item": 100,
```
estyxx, line 45: based on the [lines](https://github.com/pythonitalia/pycon/pull/2557/files#diff-06f612155123ab4a22841f0ccbef0d6aa5bd18fc876bba451f23838e6e240dd4R41-R42)
```python
        item_data = pretix_api.get_item_data(position["item"])
        category = categories.get(item_data["category"], None)
```
I would expect the "item" to be a dictionary...

marcoacierno: Item is an ID in the order

In the code you pushed you can see we run a query to get the item data (first line you copied)

association-backend/src/webhooks/handlers/pretix/pretix_event_order_paid.py
marcoacierno: The subscription will get marked as active again

association-backend/src/webhooks/handlers/pretix/pretix_event_order_paid.py
marcoacierno: This checks that we found a category from the ones we fetched

I actually changed this code so avoid fetching N+1 items, now we fetch all the items of the category so we can just do a check if the item ID exists or not in this list

association-backend/src/association/auth.py
```diff
@@ -0,0 +1,18 @@
...
+from src.association.settings import PASTAPORTO_SECRET
+from src.webhooks.auth_backend import PretixAuthBackend
+
+
+class WrapperAuthBackend(AuthenticationBackend):
+    def __init__(self):
+        self.pretix_auth_backend = PretixAuthBackend()
+        self.pastaporto_backend = PastaportoAuthBackend(PASTAPORTO_SECRET)
+
+    async def authenticate(self, request):
```
marcoacierno, line 13: doesn't seem to be a better way to handle this :( we need different auth backend based on the path, two middleware override each other

association-backend/src/webhooks/handlers/pretix/pretix_event_order_paid.py
estyxx: Ok, but here we will get `None` or something in `subscription` var? 
Because if we got an object with `status.CANCELED` for example, on line 111 the check will fail, isn't it?

marcoacierno: If the subscription is canceled it will still be found, just with status CANCELED

>  Because if we got an object with status.CANCELED for example, on line 111 the check will fail, isn't it?

Maybe we are looking at a different code?

If no subscription is found, a new one is created: https://github.com/pythonitalia/pycon/pull/2557/files/12f5633268f4b01a2b92533cdb240de8bdb7b2ca#diff-06f612155123ab4a22841f0ccbef0d6aa5bd18fc876bba451f23838e6e240dd4R101 This will happen when the user never subscribed to the association, a subscription gets created on the fly.

If a subscription is found, but is active, we raise an error: https://github.com/pythonitalia/pycon/pull/2557/files/12f5633268f4b01a2b92533cdb240de8bdb7b2ca#diff-06f612155123ab4a22841f0ccbef0d6aa5bd18fc876bba451f23838e6e240dd4R104 If that happens, we would have to refund them the 10EUR manually.

Let me know if I am looking at the wrong place, I am not 100% sure I understand what you mean here #1431

estyxx: Oh right, you don't have to recreate the object subscription again... ok then

estyxx (approved):

association-backend/src/association_membership/domain/entities.py
```diff
@@ -50,7 +50,7 @@ class Meta(BaseMeta):
         nullable=False,
     )
 
-    _add_stripe_subscription_payment: List[StripeSubscriptionPayment] = PrivateAttr(
+    _add_payments: List[Union[StripeSubscriptionPayment, PretixPayment]] = PrivateAttr(
```
patrick91, line 53: any reason to keep this as private (I mean the underscore)?

association-backend/src/association_membership/domain/repository.py
```diff
@@ -94,10 +95,8 @@ async def create_stripe_portal_session_url(
     async def save_subscription(self, subscription: Subscription) -> Subscription:
         await subscription.update()
 
-        for (
-            stripe_subscription_payment
-        ) in subscription._add_stripe_subscription_payment:
-            await stripe_subscription_payment.payment.save()
-            await stripe_subscription_payment.save()
+        for add_payment in subscription._add_payments:
```
patrick91, line 98: shall we rename `_add_payments` to `payments_to_add` or `new_payments`?

association-backend/src/association_membership/tests/domain/test_entities.py
```diff
@@ -56,8 +56,5 @@ async def _():
         invoice_pdf="https://pdfpdf",
     )
 
-    assert len(subscription._add_stripe_subscription_payment) > 0
-    assert (
-        subscription._add_stripe_subscription_payment[0].stripe_subscription_id
-        == "cs_xxx"
-    )
+    assert len(subscription._add_payments) > 0
```
patrick91, line 59: wouldn't be better to test that the len is 1? 😊

association-backend/src/webhooks/auth_backend.py
```diff
@@ -0,0 +1,38 @@
...
+    SimpleUser,
+)
+
+from src.association.settings import PRETIX_WEBHOOK_SECRET
+
+
+class PretixAuthBackend(AuthenticationBackend):
+    async def authenticate(self, request):
+        if request.url.path != "/pretix-webhook":
+            return
```
patrick91, line 17: should we raise an error just in case? :)

association-backend/src/webhooks/handlers/pretix/pretix_event_order_paid.py
```diff
@@ -0,0 +1,171 @@
...
+            f"No user found with the email of order_code={order_code}"
+        )
+
+    user_id = int(data["userByEmail"]["id"])
+    repository = AssociationMembershipRepository()
+
+    idempotency_key = PretixPayment.generate_idempotency_key(
+        organizer, event, order_code
+    )
+    if await repository.is_payment_already_processed(idempotency_key):
```
patrick91, line 97: maybe in future we can move the key generation somewhere else, right now we have it in two places (https://github.com/pythonitalia/pycon/pull/2557/files#diff-e0833ab93d303d211218ab08e427cd0d33b122aa332d261b6ee5451e99f72918R86)

patrick91 (approved):
> The rest is pretty similar, we create a Pretix payment from that as PAID, check that the payment date is within the current date (this is for safety in case we receive old events) and if it is, mark the subscription as active

did we consider also the eventuality that we might be missing some events? Maybe we can do another PR with task just for finding unsynced orders :)

other than that it looks great! 👏

association-backend/src/association_membership/tests/domain/test_entities.py
marcoacierno: sure 👍

association-backend/src/webhooks/auth_backend.py
marcoacierno: In an ideal world returning None should mean "go to the next authentication backend" so returning an error feels wrong 🤔  i.e see https://www.starlette.io/authentication/ when the header is missing

In the real world, this doesn't matter because this auth backend never gets invoked when it isn't pretix-webhook since we do the routing here: https://github.com/pythonitalia/pycon/blob/aa33ea5f78a396b7d8fe370c4c6ca6f4d2c518b7/association-backend/src/association/auth.py#L15


I would keep the `return` just because it is similar to what behaviour Starlette recommends in their doc, and if in the future they add support for multiple backends it should work out of the box (this is just guessing, since who knows what will happen in X months/years)

association-backend/src/webhooks/handlers/pretix/pretix_event_order_paid.py
marcoacierno: For now, it should be fine, and unless we want to make our entity use the repository, it isn't something we can easily fix 🤔  unless we let the DB raise an error when trying to save and not do any check before.

> maybe in future we can move the key generation somewhere else

The key itself is only built by `generate_idempotency_key` and then used in the two places (check if the payment already exists and the function where the payment is stored).

If we are worried about `add_payment` change and use a different key, we could pass it as argument

association-backend/src/association_membership/domain/entities.py
marcoacierno: Mainly because it isn't part of the domain and shouldn't be an part of the domain

association-backend/src/association_membership/domain/repository.py
marcoacierno: I can rename it `_payments_to_add` (see my answer above on why I prefer the `_`)

marcoacierno:
> did we consider also the eventuality that we might be missing some events? Maybe we can do another PR with task just for finding unsynced orders :)

Yes but I don't think it should be in this PR's scope the responsibility to also find unsynced orders 

Finding unsynced orders requires us to store somewhere the last created_at we checked, get all those created after, check and trigger this code if they are not synced so not something I would do here

association-backend/src/webhooks/auth_backend.py
marcoacierno: Changed this at the end because I want to make sure `PretixAuthBackend` is not used in some other way

PR #2564, opened 2021-12-26
Association Membership cronjob/check

marcoacierno:


association-backend/src/webhooks/handlers/crons/membership_check_status.py
```diff
@@ -0,0 +1,64 @@
...
+
+async def update_expired_subscriptions():
+    repository = AssociationMembershipRepository()
+
+    now = datetime.now(timezone.utc)
+    # Ideally in the future we should use the psql NOW() function
+    qs = Subscription.objects.filter(status=SubscriptionStatus.ACTIVE,).exclude(
+        payments__status=PaymentStatus.PAID,
+        payments__period_start__lte=now,
+        payments__period_end__gte=now,
```
estyxx, line 52: But if the payment end period is Greater Than or Equal now, doesn't mean it's still active??

marcoacierno: Yeah, that's why it is inside the `exclude`

```
qs = Subscription.objects.filter(status=SubscriptionStatus.ACTIVE,).exclude(
```

so we exclude subscriptions that are active

marcoacierno: so period_start should be in the past (lte)
and period_end should be in the future (gte)

so the subscription is still covered, exclude it

estyxx (approved):

PR #2569, opened 2021-12-28
Add filters and search fields to Admins

estyxx:

### DIFF
--- a/backend/submissions/admin.py
+++ b/backend/submissions/admin.py
@@ -89,7 +89,6 @@ class SubmissionAdmin(AdminUsersMixin):
         "abstract",
         "notes",
         "previous_talk_video",
-        "comments__text",
     )
     prepopulated_fields = {"slug": ("title",)}
     filter_horizontal = ("tags",)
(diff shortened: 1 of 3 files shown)


backend/submissions/admin.py
```diff
@@ -83,7 +83,14 @@ class SubmissionAdmin(AdminUsersMixin):
...
     )
     list_filter = ("conference", "type", "topic", "status")
-    search_fields = ("title", "abstract")
+    search_fields = (
+        "title",
+        "elevator_pitch",
+        "abstract",
+        "notes",
+        "previous_talk_video",
+        "comments__text",
```
marcoacierno, line 92: do we think it might be helpful to search the comments too?

estyxx: Maybe when you only remember "hey someone has commented to that pr about...."  ? 
I don't know tbh, I added anything that can be helpful but I should ask to the people that will use this admin...

estyxx: well I can remove it and we can add if it's needed...

marcoacierno (approved):

PR #2571, opened 2021-12-29
Add new Keynotes page and Keynote detail page

marcoacierno:
## Why

The idea is to have a separate system to store keynoters for our conference, we used the schedule item before but I feel like a new "Keynote" model might fit more. We could then connect a ScheduleItem of type keynote by adding a new FK to "keynote" like we did for submission.

I am open for discussion here in case we think using ScheduleItem is better

image

image

image

homepage

image

all keynotes page

image

the admin:

image

image

image

## How to test it

See live on Vercel:

Keynotes: https://pycon-git-web-214-keynote-page-and-homepage-python-italia.vercel.app/en/keynotes/
Keynoter page:
https://pycon-git-web-214-keynote-page-and-homepage-python-italia.vercel.app/en/keynotes/psycopg-from-2-to-3 (to translate)
https://pycon-git-web-214-keynote-page-and-homepage-python-italia.vercel.app/en/keynotes/how-the-internet-could-be-healthier

PR #2575, opened 2022-01-01
Add Job Board to PyCon Backend

patrick91:
This PR adds a Job Board, since we need as we sell it to our sponsors. I've decided to add it our main backend instead of creating a new service to keep it simple, we can make a new service in future 😊

This is based on our previous job board and how the blog posts are structured, there's some code duplication which we should remove (maybe in a future PR).

This PR only adds the backend for the job board, see tests to see how it is implemented.

As for blog posts, we should do pagination, but we can do that in future :)


backend/api/job_board/schema.py
```diff
@@ -0,0 +1,50 @@
...
+                slug=listing.slug,
+                description=listing.description,
+                company=listing.company,
+                company_logo=(
+                    info.context.request.build_absolute_uri(listing.company_logo.url)
+                    if listing.company_logo
+                    else None
+                ),
+                apply_url=listing.apply_url,
+            )
```
patrick91, line 27: do we have a new pattern for these kind of things? are we fully migrating to DDD?

marcoacierno: This is what I did in another PR: https://github.com/pythonitalia/pycon/pull/2571/files#diff-d65bfc90e5a4a3392ce87e3c0eab864a0b05b1cd7ec8a860a9803a975beba1dfR65

backend/job_board/models.py
```diff
@@ -0,0 +1,48 @@
...
+
+        filters = Q()
+
+        for lang, __ in settings.LANGUAGES:
+            filters |= Q(**{f"slug__{lang}": term})
+
+        return self.get_queryset().filter(filters)
+
+
+class JobListing(TimeStampedModel):
```
marcoacierno, line 25: should we add OrderedModel so we can order the job listings?

backend/api/tests/test_job_board.py
```diff
@@ -0,0 +1,112 @@
...
+        "description": str(listing.description),
+        "company": str(listing.company),
+        "companyLogo": get_image_url_from_request(request, listing.company_logo),
+        "applyUrl": str(listing.apply_url),
+    } == resp["data"]["jobListings"][0]
+
+
+@mark.django_db
+def test_query_single_job_listing(rf, graphql_client, job_listing_factory):
+    request = rf.get("/")
```
estyxx, line 51: `request` is not used in this test, can we remove it?

backend/api/tests/test_job_board.py
```diff
@@ -0,0 +1,112 @@
...
+            jobListing(slug: "slug") {
+                title(language: "it")
+                slug(language: "it")
+            }
+        } """
+    )
+
+    assert not resp.get("errors")
+    assert resp["data"]["jobListing"] == {
+        "title": "diventa una lumaca",
```
estyxx, line 110: 😂

backend/job_board/models.py
```diff
@@ -0,0 +1,48 @@
...
+    title = I18nCharField(_("title"), max_length=200)
+    slug = I18nCharField(_("slug"), max_length=200, blank=True)
+    company = models.CharField(_("company"), max_length=100)
+    company_logo = models.ImageField(
+        _("company logo"), null=True, blank=True, upload_to="job-listings"
+    )
+    description = I18nTextField(_("description"), blank=True)
+    apply_url = models.URLField(_("URL where you can apply"), blank=True)
+
+    objects = JobListingManager()
```
estyxx, line 35: The job offers are not related to the conference at tall? We will just delete it every year I suppose then...

backend/job_board/admin.py
```diff
@@ -0,0 +1,9 @@
+from django.contrib import admin
+
+from .models import JobListing
+
+
+@admin.register(JobListing)
+class PostAdmin(admin.ModelAdmin):
+    model = JobListing
+    list_display = ("title", "company")
```
estyxx, line 9: can we add some filters and searchers?

backend/job_board/models.py
```diff
@@ -0,0 +1,48 @@
...
+    class Meta:
+        ordering = ["created"]
+
+    def __str__(self):
+        return f"[{self.company}] - {self.title}"
+
+    def save(self, *args, **kwargs):
+        if not self.slug:
+            self.slug = copy(self.title)
+            self.slug.map(slugify)
```
estyxx, line 46:
```suggestion
            self.slug = slugify(self.title)
```
Why not just use the `slugify` django's function?

backend/job_board/models.py
patrick91: I think they should be ordered by creation date, we could add some additional flags like "highlighted" to move them at the top if needed, but we don't need custom order for now :)

backend/job_board/models.py
patrick91: I've asked Matteo, even about things like expiration and so on :)

We don't need anything complex so for now a global listing page is fine, we can think about improvements in future :D

backend/job_board/admin.py
patrick91: there's not going to be a lot of jobs :)

backend/job_board/models.py
patrick91: because slug is a dictionary :)

backend/job_board/models.py
marcoacierno: So we won't have cases where a sponsor who paid 5k wants their job to appear at the start and not below those who paid 500 euros or whatever?

Fine if we don't want to add it for now

backend/api/job_board/schema.py
```diff
@@ -0,0 +1,49 @@
...
+
+from .types import JobListing as JobListingType
+
+
+@strawberry.type
+class JobBoardQuery:
+    @strawberry.field
+    def job_listings(self, info) -> List[JobListingType]:
+        return [
+            JobListingType(
```
marcoacierno, line 14: I think we can avoid duplicating this init logic (here and in job_listing detail) with the `from_django_model`: https://github.com/pythonitalia/pycon/pull/2571/files#diff-d65bfc90e5a4a3392ce87e3c0eab864a0b05b1cd7ec8a860a9803a975beba1dfR65

backend/job_board/models.py
marcoacierno: TIL LazyI18nString has `map` :D had no idea, was confused why this worked just to find out that we have a map function in the class

backend/api/job_board/types.py
```diff
@@ -0,0 +1,35 @@
...
+
+@strawberry.type
+class JobListing:
+    id: strawberry.ID
+    title: str = strawberry.field(resolver=make_localized_resolver("title"))
+    slug: str = strawberry.field(resolver=make_localized_resolver("slug"))
+    description: str = strawberry.field(resolver=make_localized_resolver("description"))
+    company: str
+    company_logo: Optional[str]
+    apply_url: Optional[str]
```
marcoacierno, line 17:
```suggestion
    apply_url: str
```

I don't think this should be null, if it isn't specified empty string

backend/api/job_board/types.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __init__(
+        self,
+        id: str,
+        title: str,
+        slug: str,
+        description: str,
+        company: str,
+        company_logo: Optional[str],
+        apply_url: Optional[str],
```
marcoacierno, line 27:
```suggestion
        apply_url: str,
```

marcoacierno (approved):
I left a comment about using `from_django_model` so to remove the code-duplication in the strawberry type and another about a type that doesn't seem right.

Overall it looks good :)

PR #2577, opened 2022-01-02
Disable suggestions and branding on the video on homepage

pauloxnet:
Added 2 option sin the Youtube `src` attribute:

... set the modestbranding parameter value to 1 to prevent the YouTube logo from displaying in the control bar.
https://developers.google.com/youtube/player_parameters#modestbranding

... if the rel parameter is set to 0, related videos will come from the same channel as the video that was just played
https://developers.google.com/youtube/player_parameters#rel

### DIFF
--- a/frontend/src/components/youtube-lite/index.tsx
+++ b/frontend/src/components/youtube-lite/index.tsx
@@ -23,7 +23,7 @@ export const YouTubeLite: React.SFC<{ videoId: string }> = ({
         frameBorder="0"
         allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
         allowFullScreen={true}
-        src={`https://www.youtube-nocookie.com/embed/${videoId}?autoplay=1`}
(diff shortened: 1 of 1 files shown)


estyxx (approved):

marcoacierno:
Should we maybe add "loop=1" so it loops instead of asking to show other videos?

pauloxnet:
with `rel=0` the player show videos from the Python Italia channel when playback of the initial video ends.
In this way we let people to see other videos from Python Italia instead of seeing again the video they have just watched.

marcoacierno:
> with `rel=0` the player show videos from the Python Italia channel when playback of the initial video ends. In this way we let people to see other videos from Python Italia instead of seeing again the video they have just watched.

Yes, I am just not 100% sure it makes sense since the reason we have that player on the homepage is to show that video, not the youtube channel or let the user play other videos.

But I just had a look at the preview and at least it opens the video in a new tab so it is good enough for now :) as long as it doesn't play in the original container it is fine

PR #2586, opened 2022-01-06
Allow Python Italia members to vote

marcoacierno:
## Why

We want to allow people that are members of Python Italia to vote

## How to test it

I am using tests right now, I will setup staging to test

Staging https://pycon-git-web-178-enable-voting-for-python-4a3897-python-italia.vercel.app/en/voting


estyxx (approved):
LGTM!

PR #2587, opened 2022-01-07
Allow voting to people who have a ticket for previous conf

marcoacierno:


backend/voting/models/included_event.py
```diff
@@ -0,0 +1,27 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from model_utils.models import TimeStampedModel
+
+
+class IncludedEvent(TimeStampedModel):
```
marcoacierno, line 6: can you think of a better name @estyxx @patrick91 ? This model is supposed to basically store all the "additional events" that need to be checked when we want to know if the user has a ticket for other events

estyxx: `AcceptedEvent`
`SupportedEvent`
?

marcoacierno: i.e this says what events should be included in the voting checks

PR #2591, opened 2022-01-12
Show chars limit int TextArea

estyxx:
## Why

Personally, I feel bed to not knowing if I reach the limit of the Chars or if I have still some space to write the stuff... So I usually ended up copying the text, opening a python terminal and type `len("wtf...")`... :'D 

So I thought it would be nice to add a little custom component to show this...

When you reach the maximum or exceed the limit the count will turn red... 

Let me know if you like it :) 

## How to test it

CleanShot 2022-01-12 at 19 13 16
CleanShot 2022-01-12 at 19 19 22

### DIFF
--- a/frontend/src/components/cfp-form/index.tsx
+++ b/frontend/src/components/cfp-form/index.tsx
@@ -281,7 +281,6 @@ export const CfpForm: React.SFC<Props> = ({
                 {...textarea("elevatorPitch")}
                 maxLength={300}
                 rows={6}
-                limit={300}
               />
             </InputWrapper>
           </Box>
@@ -390,7 +389,6 @@ export const CfpForm: React.SFC<Props> = ({
             {...textarea("abstract")}
             rows={6}
             maxLength={5000}
-            limit={5000}
           />
         </InputWrapper>
 
--- a/frontend/src/components/cfp-form/index.tsx
+++ b/frontend/src/components/cfp-form/index.tsx
@@ -29,7 +29,7 @@ import { Alert } from "../alert";
 import { Button } from "../button/button";
 import { TagLine } from "../input-tag";
 import { InputWrapper } from "../input-wrapper";
-import LimitedTextarea from "../limited-textarea/index";
+import { LimitedTextarea } from "../limited-textarea";
 
 export type CfpFormFields = {
   type: string;
--- a/frontend/src/components/cfp-form/index.tsx
+++ b/frontend/src/components/cfp-form/index.tsx
@@ -29,6 +29,7 @@ import { Alert } from "../alert";
 import { Button } from "../button/button";
 import { TagLine } from "../input-tag";
 import { InputWrapper } from "../input-wrapper";
+import LimitedTextarea from "../limited-textarea/index";
 
 export type CfpFormFields = {
   type: string;
@@ -272,7 +273,7 @@ export const CfpForm: React.SFC<Props> = ({
               }
               errors={getErrors("validationElevatorPitch")}
             >
-              <Textarea
+              <LimitedTextarea
                 sx={{
                   resize: "vertical",
                   minHeight: 340,
@@ -280,6 +281,7 @@ export const CfpForm: React.SFC<Props> = ({
                 {...textarea("elevatorPitch")}
                 maxLength={300}
                 rows={6}
+                limit={300}
               />
             </InputWrapper>
           </Box>
--- a/frontend/src/components/cfp-form/index.tsx
+++ b/frontend/src/components/cfp-form/index.tsx
@@ -382,13 +382,15 @@ export const CfpForm: React.SFC<Props> = ({
           description={<FormattedMessage id="cfp.abstractDescription" />}
           errors={getErrors("validationAbstract")}
         >
-          <Textarea
+          <LimitedTextarea
             sx={{
               resize: "vertical",
               minHeight: 200,
             }}
             {...textarea("abstract")}
             rows={6}
+            maxLength={5000}
+            limit={5000}
           />
         </InputWrapper>
 
--- a/frontend/src/components/limited-textarea/index.tsx
+++ b/frontend/src/components/limited-textarea/index.tsx
@@ -1,26 +1,21 @@
 import { useState } from "react";
 import { Text, Textarea, TextareaProps } from "theme-ui";
 
-type LimitedTextareaPros = TextareaProps & {
-  value: string;
-  limit: number;
-};
-
 export const LimitedTextarea = ({
   value,
-  limit,
+  maxLength,
   ...props
-}: LimitedTextareaPros) => {
+}: TextareaProps) => {
   return (
     <>
-      <Textarea value={value} {...props} />
+      <Textarea value={value} maxLength={maxLength} {...props} />
       <Text
         variant="labelDescription"
         as="p"
         mb={4}
-        color={value.length >= limit ? "red" : "black"}
+        color={(value as string)?.length >= maxLength ? "red" : "black"}
       >
-        {value.length}/{limit}
+        {(value as string)?.length}/{maxLength}
       </Text>
     </>
   );
--- a/frontend/src/components/limited-textarea/index.tsx
+++ b/frontend/src/components/limited-textarea/index.tsx
@@ -1,11 +1,13 @@
-import { useState } from "react";
-import { Text, Textarea, TextareaProps } from "theme-ui";
+import { Text, Textarea } from "theme-ui";
 
 export const LimitedTextarea = ({
   value,
   maxLength,
   ...props
-}: TextareaProps) => {
+}: {
+  value: string;
+  maxLength: number;
+}) => {
   return (
     <>
       <Textarea value={value} maxLength={maxLength} {...props} />
@@ -13,11 +15,10 @@ export const LimitedTextarea = ({
         variant="labelDescription"
         as="p"
         mb={4}
-        color={value?.length >= maxLength ? "red" : "black"}
+        color={value.length >= maxLength ? "red" : "black"}
       >
-        {value?.length}/{maxLength}
+        {value.length}/{maxLength}
       </Text>
     </>
   );
 };
-
--- a/frontend/src/components/limited-textarea/index.tsx
+++ b/frontend/src/components/limited-textarea/index.tsx
@@ -12,13 +12,10 @@ const LimitedTextarea = ({
 }: LimitedTextareaPros) => {
   const [content, setContent] = useState(value.slice(0, limit));
 
-  console.log(value, content);
-
   return (
     <>
       <Textarea
         onChange={(event) => {
-          console.log(event.target.value);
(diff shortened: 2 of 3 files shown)


marcoacierno:
tslint is failing? https://github.com/pythonitalia/pycon/runs/4794294611?check_suite_focus=true

frontend/src/components/cfp-form/index.tsx
```diff
@@ -29,6 +29,7 @@ import { Alert } from "../alert";
 import { Button } from "../button/button";
 import { TagLine } from "../input-tag";
 import { InputWrapper } from "../input-wrapper";
+import LimitedTextarea from "../limited-textarea/index";
```
marcoacierno, line 32:
```suggestion
import LimitedTextarea from "../limited-textarea";
```

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+        mb={4}
+        color={value.length >= limit ? "red" : "black"}
+      >
+        {content.length}/{limit}
+      </Text>
+    </>
+  );
+};
+
+export default LimitedTextarea;
```
marcoacierno, line 37:
```suggestion

```

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,37 @@
+import { useState } from "react";
+import { Text, Textarea, TextareaProps } from "theme-ui";
+
+type LimitedTextareaPros = TextareaProps & {
+  limit: number;
+};
+const LimitedTextarea = ({
```
marcoacierno, line 7: we usually prefer export naming constants

```suggestion
export const LimitedTextarea = ({
```

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+  value,
+  limit,
+  onChange,
+  ...props
+}: LimitedTextareaPros) => {
+  const [content, setContent] = useState(value.slice(0, limit));
+
+  return (
+    <>
+      <Textarea
```
marcoacierno, line 17: browsers have a `maxlength` property to automatically handle this, woudln't it work better than changing the content like this?

frontend/src/components/cfp-form/index.tsx
```diff
@@ -272,14 +273,15 @@ export const CfpForm: React.SFC<Props> = ({
...
-              <Textarea
+              <LimitedTextarea
                 sx={{
                   resize: "vertical",
                   minHeight: 340,
                 }}
                 {...textarea("elevatorPitch")}
                 maxLength={300}
                 rows={6}
+                limit={300}
```
marcoacierno, line 284: do we have other places where we have a limit of chars? might be good to add it everywhere we have limits

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+        onChange={(event) => {
+          onChange(event);
+          setContent(event.target.value.slice(0, limit));
```
marcoacierno, line 20: I don't think this is correct

the local Content will be updated to have the "sliced" version, but we will still send the long value to the component upstream (the CFP Form) because we do `onChange(event)` which will take the content from `event`, so when sending the form to GraphQL you would still send the long value causing an error.

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+        value={value}
+        {...props}
+      />
+      <Text
+        variant="labelDescription"
+        as="p"
+        mb={4}
+        color={value.length >= limit ? "red" : "black"}
+      >
+        {content.length}/{limit}
```
marcoacierno, line 31:
```suggestion
        {value.length}/{limit}
```

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <>
+      <Textarea
+        onChange={(event) => {
+          onChange(event);
+          setContent(event.target.value.slice(0, limit));
+        }}
+        value={value}
+        {...props}
+      />
+      <Text
```
marcoacierno, line 25:
```suggestion
        }}
        value={value}
        {...props}
        maxLength={limit}
      />
      <Text
```

estyxx:
```Error: src/components/limited-textarea/index.tsx(15,8): error TS2322: Type '{ ref?: LegacyRef<HTMLTextAreaElement>; key?: Key; autoComplete?: string; autoFocus?: boolean; cols?: number; dirName?: string; disabled?: boolean; form?: string; ... 296 more ...; value: (string | ... 1 more ... | readonly string[]) & string; }' is not assignable to type 'RefAttributes<HTMLTextAreaElement>'.
``` 
I have no idea how to fix this :( @marcoacierno

frontend/src/components/cfp-form/index.tsx
```diff
@@ -380,13 +382,15 @@ export const CfpForm: React.SFC<Props> = ({
...
-          <Textarea
+          <LimitedTextarea
             sx={{
               resize: "vertical",
               minHeight: 200,
             }}
             {...textarea("abstract")}
             rows={6}
+            maxLength={5000}
+            limit={5000}
```
marcoacierno, line 393: we already pass `limit` to `LimitedTextarea`, why do we have to specify it twice? Why doesn't `LimitedTextarea` do it? In what case do we want to show a limit and enforce another?

```suggestion
            limit={5000}
```

I would prefer if the LimitedTextarea was responsible for enforcing it so we are consistent

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+  limit: number;
+};
+
+export const LimitedTextarea = ({
+  value,
+  limit,
+  ...props
+}: LimitedTextareaPros) => {
+  return (
+    <>
```
marcoacierno, line 14: I wonder if for accessibility reason the wrapper should be a label 🤔

frontend/src/components/submission-comments/add-comment.tsx
```diff
@@ -62,14 +63,15 @@ export const AddComment: React.SFC<Props> = ({ submissionId }) => {
   return (
     <Box as="form" onSubmit={onSubmit} sx={{ mb: 4 }}>
       <InputWrapper sx={{ mb: 2 }} errors={getErrors("validationText")}>
-        <Textarea
+        <LimitedTextarea
```
marcoacierno, line 66: we don't use Textarea anymore, so we can remove the import

frontend/src/components/limited-textarea/index.tsx
estyxx: I forgot about accessibility :( 
We should definitely have a look at it!

frontend/src/components/cfp-form/index.tsx
estyxx: Yeah agree to pass only one of them, I will keep `maxLength` only as it's already used for TextArea

marcoacierno: 👍  let's replace limit with maxLength then

marcoacierno:
> ```
> Error: src/components/limited-textarea/index.tsx(15,8): error TS2322: Type '{ ref?: LegacyRef<HTMLTextAreaElement>; key?: Key; autoComplete?: string; autoFocus?: boolean; cols?: number; dirName?: string; disabled?: boolean; form?: string; ... 296 more ...; value: (string | ... 1 more ... | readonly string[]) & string; }' is not assignable to type 'RefAttributes<HTMLTextAreaElement>'.
> ```
> 
> I have no idea how to fix this :( @marcoacierno

let me know when you pushed all changes and I will have a look 👍

estyxx:
> > ```
> > Error: src/components/limited-textarea/index.tsx(15,8): error TS2322: Type '{ ref?: LegacyRef<HTMLTextAreaElement>; key?: Key; autoComplete?: string; autoFocus?: boolean; cols?: number; dirName?: string; disabled?: boolean; form?: string; ... 296 more ...; value: (string | ... 1 more ... | readonly string[]) & string; }' is not assignable to type 'RefAttributes<HTMLTextAreaElement>'.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I have no idea how to fix this :( @marcoacierno
> 
> let me know when you pushed all changes and I will have a look 👍

I used only `TextareaProps` from the library now without merging the two types but is still there... wondering if it something in `theme-ui` that should be fixed 🤔

frontend/src/components/submission-comments/add-comment.tsx
marcoacierno: @estyxx this is still not solved. You can see above there is the eslint warning about the import

frontend/src/components/limited-textarea/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+      <Textarea value={value} maxLength={maxLength} {...props} />
+      <Text
+        variant="labelDescription"
+        as="p"
+        mb={4}
+        color={(value as string)?.length >= maxLength ? "red" : "black"}
+      >
+        {(value as string)?.length}/{maxLength}
```
marcoacierno, line 17: What types can `value` be? also using `value?.` doesn't seem ideal since this will throw an error because it will end up being `undefined.length`?

marcoacierno: can we maybe change the value type so it is always a string?

frontend/src/components/limited-textarea/index.tsx
estyxx: I can put something like this in the Text
` aria-label={`${(value as string)?.length} characters left over ${maxLength}`}` 
but we need much more than that... we don't use any `aria-label` (don't know if there are other fields to improve  accessibility...) 
We can work on that in another issue I think

frontend/src/components/limited-textarea/index.tsx
estyxx: Typescript was red because it can be null | string | string[]

estyxx: can you check again? I set value to string now :)

estyxx: No typescript complains again... 
> Error: src/components/limited-textarea/index.tsx(13,8): error TS2322: Type '{ ref?: LegacyRef<HTMLTextAreaElement>; key?: Key; autoComplete?: string; autoFocus?: boolean; cols?: number; dirName?: string; disabled?: boolean; form?: string; ... 296 more ...; maxLength: number; }' is not assignable to type 'RefAttributes<HTMLTextAreaElement>'.

But I don't know how to solve this... :(

marcoacierno:
rebased & cleaned up this PR here: #3015

PR #2597, opened 2022-01-15
Improve Submission admin

estyxx:
## Why

- Be able to search by User name/email / full name 
- Open the submission on pycon.it website


backend/users/client.py
```diff
@@ -16,6 +16,14 @@
 }
 """
 
+GET_USERS_BY_NAME = """
```
marcoacierno, line 19: I would name it more specific to what it does so 🤔 

```suggestion
SEARCH_USERS = """
```

marcoacierno (approved):

PR #2599, opened 2022-01-17
Improve UI when user cannot vote

estyxx:
## Why

When the page is loading `isVotingClosed` is undefined so we were showing for a second the filters, was a bit ugly. 

Also adjust the margin for errors to be equal to the first section..

## How to test it

CleanShot 2022-01-17 at 13 02 58
CleanShot 2022-01-17 at 13 02 36
CleanShot 2022-01-17 at 13 13 17

### DIFF
--- a/frontend/src/locale/index.ts
+++ b/frontend/src/locale/index.ts
@@ -411,6 +411,7 @@ export const messages = {
     "voting.errors.cannotVote.body":
       "Please check our {linkVotingInfo} and buy your ticket {linkTicket} to join us at PyCon Italia",
     "voting.errors.cannotVote.linkVotingInfo.text": "voting guidelines",
+    "voting.errors.cannotVote.linkTicket.text": "here",
     "voting.loading": "Loading",
     "voting.fullDetails": "Full details",
     "voting.allTopics": "All topics",
@@ -856,8 +857,9 @@ export const messages = {
     "voting.needToBeLoggedIn": "Esegui l'accesso o crea un account per votare",
     "voting.errors.cannotVote.heading": "Ci dispiace, non puoi votare!",
     "voting.errors.cannotVote.body":
-      "Per favore, controlla le nostre {linkVotingInfo} per votare e acquista il tuo biglietto {linkTicket} per partecipare a PyCon Italia",
+      "Controlla le nostre {linkVotingInfo} per votare e acquista il tuo biglietto {linkTicket} per partecipare a PyCon Italia",
     "voting.errors.cannotVote.linkVotingInfo.text": "linee guida",
+    "voting.errors.cannotVote.linkTicket.text": "qui",
     "voting.fullDetails": "Continua a leggere",
     "voting.allTopics": "Tutti gli argomenti",
     "voting.allLanguages": "Tutte le lingue",
--- a/frontend/src/locale/index.ts
+++ b/frontend/src/locale/index.ts
@@ -5,9 +5,13 @@ export const messages = {
       "Join PyCon Italia from the 2nd to the 5th of June 2022 in Florence, Italy!",
     "global.yes": "Yes",
     "global.no": "No",
+    "global.learnMore": "Learn more",
     "global.loading": "Loading",
     "global.button.loading": "Loading {emoji}",
     "global.somethingWentWrong": "Something went wrong, please try again",
+    "global.accordion.close": "Close",
+    "global.accordion.readMore": "Read more",
+
     "home.title": "Home",
     "home.deadline.begins": "Begins",
     "home.deadline.deadline": "Deadline",
@@ -118,7 +122,6 @@ export const messages = {
       "PyCon Italia is seeking speakers of all experience levels and backgrounds to contribute to our conference program! If you use Python professionally, as a hobbyist or are just excited about Python or programming and open source communities, we’d love to hear from you. We want you and your ideas at PyCon Italia!",
     "cfp.introductionDeadline":
       "You have until {deadline} to submit a proposal!",
-    "cfp.learnMore": "Learn more",
     "cfp.youridea": "Your idea",
     "cfp.choosetype": "Choose a format",
     "cfp.format": "Format",
@@ -285,7 +288,6 @@ export const messages = {
       "Thanks to the help of the Python Software Foundation we are glad to give financial support to everyone who may need it in order to join PyCon Italia.",
     "grants.introductionDeadline":
       "You have until {deadline} to send your request!",
-    "grants.learnMore": "Learn more",
     "grants.comingSoon": "The grants form will open on {start}!",
     "grants.closed":
       "The grants form is now closed! If you sent a request, you will be contacted by us in the next few weeks with our decision.",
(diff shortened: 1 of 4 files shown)


frontend/src/locale/index.ts
```diff
@@ -852,7 +853,9 @@ export const messages = {
     "voting.sendingVote": "Invio in corso",
     "voting.voteSent": "Voto inviato!",
     "voting.needToBeLoggedIn": "Esegui l'accesso o crea un account per votare",
-    "voting.buyTicketToVote": "Devi comprare un biglietto per poter votare",
+    "voting.errors.cannotVote.heading": "Ci dispiace, non puoi votare!",
+    "voting.errors.cannotVote.body":
+      "Per favore, controlla le nostre linee guida per votare su {linkVotingInfo} e acquista il tuo biglietto {linkTicket} per partecipare a PyCon Italia",
```
marcoacierno, line 858:
```suggestion
    "voting.errors.cannotVote.body":
      "Controlla le nostre linee guida per votare su {linkVotingInfo} e acquista il tuo biglietto {linkTicket} per partecipare a PyCon Italia",
```

frontend/src/pages/voting/index.tsx
```diff
@@ -104,6 +104,8 @@ export const VotingPage = () => {
     ) !== -1;
 
   const isVotingClosed = data && !data.conference.isVotingOpen;
+  const userCannotVote =
+    loggedIn && (loading || (cannotVoteErrors ?? false) || (error ?? false));
```
marcoacierno, line 108: This line looks very complex

Can you describe in words what you are trying to achieve?

frontend/src/pages/voting/index.tsx
```diff
@@ -130,108 +132,127 @@ export const VotingPage = () => {
...
-                <FormattedMessage id="voting.buyTicketToVote" />
-              </Link>
-            </Alert>
-          )}
-          {loading && (
-            <Alert variant="info">
-              <FormattedMessage id="voting.loading" />
-            </Alert>
-          )}
+      {userCannotVote && (
```
marcoacierno, line 218: Uhm this looks a bit weird to me / complex, the nesting is not ideal also I don't think we should show a "You can't vote" while the network is loading (this is what happens right?), it is very confusing

frontend/src/pages/voting/index.tsx
```diff
@@ -130,108 +132,127 @@ export const VotingPage = () => {
...
-                    <option key={language.id} value={language.code}>
-                      {language.name}
-                    </option>
-                  ))}
-                </Select>
-
-                <Select
-                  {...select("vote")}
+            {isVotingClosed !== undefined &&
+              !isVotingClosed &&
```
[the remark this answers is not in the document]
estyxx: Yes because when there is an error (user cannot vote) isVotingClosed is undefined anyway (but the voting is open actually... ) so setting it as default Close was not right...

frontend/src/pages/voting/index.tsx
estyxx: I would like to reuse this condition to understand whether I have to show the filters or not... In order to not have the ugly thing that the filter show and disappear immediately when user cannot vote

frontend/src/pages/voting/index.tsx
estyxx: Nope when the network is loading we will show Loading, because the above condition at line 230

frontend/src/pages/voting/index.tsx
estyxx: Agree is quite complex, took me some time to figure out the right condition... :D

marcoacierno: So I am having a look

cannotVoteErrors:

can be changed to

```typescript
  const cannotVoteErrors =
    (error?.graphQLErrors ?? []).findIndex(
      (e) => e.message === "You need to have a ticket to see submissions",
    ) !== -1;
```

so you don't need `(cannotVoteErrors ?? false)`, cannotVoteErrors will always be true.

Note that I think this would have worked without `??` maybe by doing `!!cannotVoteErrors` so it goes from undefined to boolean :)

marcoacierno: Another thing, why do we `error ?? false`? why do we check error here? We already kind of do it in `cannotVoteErrors` and if I have to be honest, if the error the user is seeing is not that they can't vote, I wouldn't want to display this message

frontend/src/pages/voting/index.tsx
```diff
@@ -104,6 +104,10 @@ export const VotingPage = () => {
     ) !== -1;
 
   const isVotingClosed = data && !data.conference.isVotingOpen;
```
marcoacierno, line 106: we could use ts here to always have a boolean?

```
const isVotingClosed = !(data?.conference?.isVotingOpen ?? false);
```

frontend/src/pages/voting/index.tsx
```diff
@@ -104,6 +104,10 @@ export const VotingPage = () => {
...
+  const userCannotVote =
+    loggedIn && (loading || (cannotVoteErrors ?? false) || (error ?? false));
+  const showFilters =
+    isVotingClosed !== undefined && !isVotingClosed && !userCannotVote;
```
marcoacierno, line 110: if we make `isVotingClosed` always a boolean this can just be

```suggestion
  const showFilters = !isVotingClosed && !userCannotVote;
```

frontend/src/pages/voting/index.tsx
marcoacierno: Oki, I see why we do `error ?? false` because we want this variable to be false when the error is not related to the voting permissions

This feels very confusing  🤔

frontend/src/pages/voting/index.tsx
estyxx: seems working :)

marcoacierno:
image


I think something went wrong with the changes?

estyxx:
> 
> I think something went wrong with the changes?

Yeah think it's not correct to set the default to `false` when the `data` is `undefined` otherwise we have to do more condition to check if the user is not logged in to decide if we want to show the closed voting message... 

Would be better if the backend return the `isVotingClosed` anyway... but even in that case we have 1 second when the page is loading when the `data` is still `undefined`

estyxx:
@marcoacierno I reverted the change, for now. I can look at it tomorrow morning to see if I can come up with something better...

frontend/src/pages/voting/index.tsx
```diff
@@ -208,30 +214,52 @@ export const VotingPage = () => {
...
+                    id="voting.errors.cannotVote.body"
+                    values={{
+                      linkVotingInfo: (
+                        <Link path="/voting-info">
+                          <FormattedMessage id="voting.errors.cannotVote.linkVotingInfo.text" />
+                        </Link>
+                      ),
+                      linkTicket: (
+                        <Link path="/tickets">
+                          {" "}
```
marcoacierno, line 248: this {" "} space looks terrible. Can we add it to the string in the locale?

```suggestion
```

marcoacierno (approved):
We can merge this for now since the message when you can't vote is better but the code is a real mess :(( we should refactor this in the future

frontend/src/pages/voting/index.tsx
estyxx: Was a mistake

PR #2600, opened 2022-01-17
Fix error when changing language in the tickets review

marcoacierno:
## Why

The hack to add language support caused some issues with lookup tables and so on. This removes it and uses a better solution

### DIFF
--- a/frontend/src/components/tickets-page/review/create-order-buttons.tsx
+++ b/frontend/src/components/tickets-page/review/create-order-buttons.tsx
@@ -28,7 +28,7 @@ export const CreateOrderButtons: React.SFC<Props> = ({ state, email }) => {
           return;
         }
 
-        window.localStorage.removeItem("tickets-cart");
+        window.localStorage.removeItem("tickets-cart-v2");
         window.location.href = result.createOrder.paymentUrl;
       },
     });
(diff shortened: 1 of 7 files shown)


frontend/src/components/tickets-page/review/create-order-buttons.tsx
```diff
@@ -28,7 +28,7 @@ export const CreateOrderButtons: React.SFC<Props> = ({ state, email }) => {
           return;
         }
 
-        window.localStorage.removeItem("tickets-cart");
+        window.localStorage.removeItem("tickets-cart-v2");
```
marcoacierno, line 31: this forces a new cart

frontend/src/components/tickets-page/review/create-order-buttons.tsx
```diff
@@ -40,7 +40,7 @@ export const CreateOrderButtons: React.SFC<Props> = ({ state, email }) => {
       )
         .flat()
         .map((product) => ({
-          ticketId: product.id.replace("-en", "").replace("-it", ""),
+          ticketId: product.id,
```
estyxx, line 43: shouldn't add the language here?

marcoacierno: this is the mapping to send the ticket id to createOrder, we only care about the ID you are purchasing, nothing else

estyxx (approved):

PR #2601, opened 2022-01-17
Add voting CTA on the homepage

marcoacierno:


estyxx (approved):
We need something dynamic asap! :D

PR #2602, opened 2022-01-18
Lazy load submissions via scroll

marcoacierno:

### DIFF
--- a/backend/api/submissions/schema.py
+++ b/backend/api/submissions/schema.py
@@ -1,7 +1,10 @@
 import typing
 
 import strawberry
-from api.permissions import HasTokenPermission
+
+from api.helpers.ids import decode_hashid
+from api.permissions import CanSeeSubmissions, IsAuthenticated
+from conferences.models import Conference as ConferenceModel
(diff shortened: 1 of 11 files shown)


backend/api/submissions/schema.py
```diff
@@ -17,9 +20,26 @@ def submission(self, info, id: strawberry.ID) -> typing.Optional[Submission]:
...
 
-    @strawberry.field(permission_classes=[HasTokenPermission])
-    def submissions(self, info, code: str) -> typing.Optional[typing.List[Submission]]:
-        return SubmissionModel.objects.filter(conference__code=code).all()
+    @strawberry.field(permission_classes=[IsAuthenticated])
+    def submissions(
+        self,
+        info,
+        code: str,
+        after: typing.Optional[str] = None,
```
estyxx, line 28: Don't we have a pagination type?

estyxx (approved):

backend/api/submissions/schema.py
marcoacierno: Do we? can you link it to me?

PR #2603, opened 2022-01-18
Move Voting filters to the BE

marcoacierno:
- Lazy load submissions via scroll
- Move filters to the BE


marcoacierno:
this was done in #3243

PR #2623, opened 2022-01-24
Filter submission by state Proposed

estyxx:

### DIFF
--- a/backend/api/conferences/types.py
+++ b/backend/api/conferences/types.py
@@ -19,6 +19,7 @@ from api.voting.types import RankSubmission
 from cms.models import GenericCopy
 from conferences.models.deadline import DeadlineStatus
 from schedule.models import ScheduleItem as ScheduleItemModel
+from submissions.models import Submission as SubmissionModel
 from voting.models import RankRequest as RankRequestModel
 
 from ..helpers.i18n import make_localized_resolver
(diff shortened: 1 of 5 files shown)


backend/api/conferences/types.py
```diff
@@ -254,8 +254,10 @@ def submission_types(self, info) -> List[SubmissionType]:
         return self.submission_types.all()
 
     @strawberry.field(permission_classes=[CanSeeSubmissions])
-    def submissions(self, info) -> Optional[List[Submission]]:
-        return self.submissions.all().select_related(
+    def submissions(
+        self, info, status: Optional[str] = None
```
marcoacierno, line 258: any reason we want to allow the clients to filter them by status?

also does this code work if status is not passed? since it would filter for `status=None` I think?

marcoacierno: why don't we just return those that are of status "proposed"?

estyxx: I thought this will be more generic. But now we have the voting and the schedule (the admin tool to select them...) the page that is querying the conference's submissions. So think I can remove the filtering and return only the Proposed submissions always

backend/api/tests/schema/conference/test_conference.py
```diff
@@ -722,3 +722,27 @@ def test_get_conference_voucher_with_invalid_code(graphql_client, conference, mo
...
+        }
+    """
+
+    response = graphql_client.query(
+        query,
+        variables={"code": conference.code},
+    )
+
+    assert len(response["data"]["conference"]["submissions"]) == 1
+    assert response["data"]["conference"]["submissions"]["status"] == "proposed"
```
marcoacierno, line 748: how does this test work? you don't select the field in the query

estyxx: indeed it was failing :D

marcoacierno (approved):
Should be good to merge :)

PR #2631, opened 2022-01-31
Add ranking page

estyxx:
## How to test it

CleanShot 2022-02-06 at 19 05 21@2x


marcoacierno:
image

Something going wrong? https://pycon-git-add-ranking-page-python-italia.vercel.app/en/ranking

estyxx:
> image
> 
> Something going wrong? [pycon-git-add-ranking-page-python-italia.vercel.app/en/ranking](https://pycon-git-add-ranking-page-python-italia.vercel.app/en/ranking)
CleanShot 2022-02-01 at 09 21 22
Weird locally I can see the details... 🤔

backend/api/submissions/permissions.py
```diff
@@ -14,6 +15,12 @@ def has_permission(self, source, info):
 
         conference = source.conference
 
+        try:
+            if conference.rankrequest:
+                return True
+        except RankRequest.DoesNotExist:
+            pass
```
marcoacierno, line 22: I wonder if there is a way to make this code look a bit btter

frontend/src/pages/ranking/ranking-submissions.graphql
```diff
@@ -14,6 +14,9 @@ query RankingSubmission($conference: String!) {
 
       submission {
         ...submissionAccordion
+        speaker {
+          fullName
+        }
```
marcoacierno, line 19:
```suggestion
        speaker {
          id
          fullName
        }
```

always fetch the ID so Apollo can cache

marcoacierno, line 19: can we select the ID as well? so apollo can cache

frontend/src/pages/ranking/index.tsx
```diff
@@ -23,7 +26,7 @@ const COLORS = [
   },
 ];
 
-export const RankingPage: React.SFC = () => {
+export const RankingPage: React.FC = () => {
```
marcoacierno, line 29:
```suggestion
export const RankingPage = () => {
```

frontend/src/pages/ranking/index.tsx
```diff
@@ -32,7 +35,15 @@ export const RankingPage: React.SFC = () => {
   });
 
   const [filters, { select }] = useFormState();
-
+  const filterVisibleSubmissions = (submission) => {
```
marcoacierno, line 38: we can keep this as it is but when we work on the voting improvements (i.e server-side filtering) we should migrate  update this page too

backend/api/submissions/permissions.py
```diff
@@ -3,6 +3,7 @@
 from api.permissions import HasTokenPermission
 from submissions.models import Submission
 from voting.helpers import pastaporto_user_info_can_vote
+from voting.models.ranking import RankRequest
 
 
 class CanSeeSubmissionRestrictedFields(BasePermission):
```
marcoacierno, line 9: not for this PR but something I think we should check if we have time: How do we handle cancelled submissions permission wise? what happens if I query a cancelled submission directly by doing `submission(id: XX)`? Ideally it should fail with a 404

marcoacierno (approved):
haven't tested, code looks good

backend/api/submissions/permissions.py
estyxx: Good point

estyxx: yeah they are visible!

backend/api/submissions/permissions.py
estyxx: don't think so, if the rank request doesn't exist it will throw the exception...

backend/api/submissions/permissions.py
estyxx: We should change that only the Author or an admin can see!

backend/voting/models/ranking.py
```diff
@@ -26,6 +27,11 @@ def __str__(self):
     def save(self, *args, **kwargs):
         super(RankRequest, self).save(*args, **kwargs)
 
+        exising_submissions = self.rank_submissions.all()
+
+        # do not recreate ranking
+        if exising_submissions:
+            return
```
marcoacierno, line 34: Uhm, this feels like it will fetch the objects just to check if it is empty or not?


```suggestion
        # do not recreate ranking
        if self.rank_submissions.exists():
            return
```

does this work? so it just checks with `SELECT EXISTS` and be more performant?

PR #2652, opened 2022-02-07
Add Ranking stats

estyxx:
## How to test it

CleanShot 2022-02-11 at 22 40 53
CleanShot 2022-02-11 at 22 40 47
CleanShot 2022-02-11 at 22 40 34
CleanShot 2022-02-11 at 22 40 24


backend/voting/models/ranking.py
```diff
@@ -6,7 +6,33 @@
...
+    class Type(models.TextChoices):
+        SPEAKERS = "speakers", _("Speakers")
+        SUBMISSIONS = "submissions", _("Submissions")
+        GENDER = "gender", _("Gender")
+        SUBMISSION_TYPE = "submission_type", _("Submission  Type")
+        LANGUAGE = "language", _("Language")
+        AUDIENCE_LEVEL = "audience_level", _("Audience Level")
+
+    # name = I18nCharField(_("name"), max_length=100)
+    name = models.CharField(_("Name"), max_length=50)
```
marcoacierno, line 24: why have a name and a type?

backend/voting/models/ranking.py
```diff
@@ -129,6 +156,81 @@ def group_by(item):
...
+            )
+
+        # N. submissions by Audience Level
+        for audience_level in self.conference.audience_levels.all():
+            count = submissions.filter(
+                submission__audience_level=audience_level
+            ).count()
+
+            RankStat.objects.create(
+                name=f"{audience_level.name}",
```
marcoacierno, line 218: maybe name is ok if you use it like this

docker-compose.yml
```diff
@@ -122,6 +122,8 @@ services:
       test: ["CMD-SHELL", "pg_isready -U users-backend"]
       interval: 5s
       timeout: 5s
+    ports:
+      - 5435:5432
```
marcoacierno, line 126:
```suggestion
```

backend/api/voting/types.py
```diff
@@ -31,15 +31,27 @@ class VoteType:
...
+
 @strawberry.type
 class RankRequest:
     is_public: bool
     ranked_submissions: List[RankSubmission]
+    stats: List[RankStat]
+
+    @strawberry.field
+    def stat(self, info, type: str) -> Optional[RankStat]:
+        return self.stats.filter(type=type).first()
```
marcoacierno, line 50: I am not sure how this works

Below you create multiple RankStat with the same type, but here you can only return one?

In your code we could have multiple of type `GENDER`

backend/api/voting/types.py
```diff
@@ -31,15 +31,27 @@ class VoteType:
...
+
+
 @strawberry.type
 class RankRequest:
     is_public: bool
     ranked_submissions: List[RankSubmission]
+    stats: List[RankStat]
+
+    @strawberry.field
+    def stat(self, info, type: str) -> Optional[RankStat]:
```
marcoacierno, line 49: if we don't use this we can remove it

frontend/src/locale/index.ts
```diff
@@ -833,9 +847,25 @@ export const messages = {
...
+`,
+    "ranking.topicComment":
+      "Abbiamo ricevuto {value} proposte per la track {name} quest'anno!",
     "ranking.seoTitle": "Ranking",
     "ranking.details": "Dettagli",
+    "ranking.stats.heading": "Alcune statistiche",
+    "ranking.stats.submissions": "{value} proposte in totale",
+    "ranking.stats.speakers": "{value} persone si sono proposte come speaker",
+    "ranking.stats.gender.women": "{value} proposte da speaker donne",
+    "ranking.stats.gender.men": "{value} proposte da speaker uomini",
```
marcoacierno, line 863: don't we also have other / prefer not to say? we should probably show those as well as "from unspecified genders" or some similar copy

estyxx: Something like:
 "{value} proposte da speaker di gender non specificato:
 "{value} proposals from unspecified gender speakers"
 ?
 
 And do we want to sum `other` and `prefer not to say` ?

backend/api/conferences/types.py
```diff
@@ -311,28 +311,26 @@ def talk(self, info, slug: str) -> Optional[ScheduleItem]:
...
-            return RankRequest(
-                is_public=rank_request.is_public,
-                ranked_submissions=submissions,
-            )
-        except RankRequestModel.DoesNotExist:
+        rank_requests = RankRequestModel.objects.filter(conference=self)
+        if not rank_requests:
+            return None
+
+        rank_request = rank_requests[0]
```
marcoacierno, line 318:
```suggestion
        rank_request = RankRequestModel.objects.filter(conference=self).first()
        if not rank_request:
            return None
```

looks a bit better and should be the same?

backend/api/tests/schema/conference/test_ranking.py
```diff
@@ -1,23 +1,30 @@
-from pytest import mark
+import pytest
 
-pytestmark = mark.django_db
+pytestmark = [pytest.mark.django_db]
 
 
-def test_conference_ranking_does_not_exists(conference_factory, graphql_client):
+@pytest.fixture
+def mock_users(mocker):
```
marcoacierno, line 7: uhm ideally we should mock with this `respx` 🤔

estyxx: I switched to `respx`

marcoacierno (approved):
looks good to me

can you write me to on telegram when you merge this? since there are migrations I fear I need to manually run the migrate script for now until I fix the workflow

PR #2657, opened 2022-02-08
Filter Ranking by topics

estyxx:

### DIFF
--- a/backend/api/conferences/types.py
+++ b/backend/api/conferences/types.py
@@ -324,7 +324,7 @@ class Conference:
                 return None
 
             submissions = rank_request.rank_submissions.filter(
-                submission__topic__id=int(topic)
+                submission__topic__id=topic
             ).order_by("absolute_rank")
             return RankRequest(
                 is_public=rank_request.is_public,
--- a/backend/api/conferences/types.py
+++ b/backend/api/conferences/types.py
@@ -310,7 +310,7 @@ class Conference:
         return self.schedule_items.filter(slug=slug).first()
 
     @strawberry.field
-    def ranking(self, info, topic: Optional[str] = None) -> Optional[RankRequest]:
+    def ranking(self, info, topic: str) -> Optional[RankRequest]:
         try:
             rank_requests = RankRequestModel.objects.filter(conference=self)
             if not rank_requests:
@@ -323,11 +323,12 @@ class Conference:
             ):
                 return None
 
+            submissions = rank_request.rank_submissions.filter(
+                submission__topic=topic
+            ).order_by("absolute_rank")
             return RankRequest(
                 is_public=rank_request.is_public,
-                ranked_submissions=rank_request.rank_submissions.filter(
-                    topic=topic
-                ).order_by("absolute_rank"),
+                ranked_submissions=submissions,
             )
         except RankRequestModel.DoesNotExist:
             return None
--- a/backend/api/conferences/types.py
+++ b/backend/api/conferences/types.py
@@ -310,7 +310,7 @@ class Conference:
         return self.schedule_items.filter(slug=slug).first()
 
     @strawberry.field
-    def ranking(self, info, topic: str) -> Optional[RankRequest]:
+    def ranking(self, info, topic_id: int) -> Optional[RankRequest]:
         try:
             rank_requests = RankRequestModel.objects.filter(conference=self)
             if not rank_requests:
@@ -324,7 +324,7 @@ class Conference:
                 return None
 
             submissions = rank_request.rank_submissions.filter(
-                submission__topic=topic
+                submission__topic__id=topic_id
             ).order_by("absolute_rank")
             return RankRequest(
                 is_public=rank_request.is_public,
--- a/backend/api/conferences/types.py
+++ b/backend/api/conferences/types.py
@@ -310,7 +310,7 @@ class Conference:
         return self.schedule_items.filter(slug=slug).first()
 
     @strawberry.field
-    def ranking(self, info, topic: str) -> Optional[RankRequest]:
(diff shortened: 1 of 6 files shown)


backend/api/conferences/types.py
```diff
@@ -310,7 +310,7 @@ def talk(self, info, slug: str) -> Optional[ScheduleItem]:
         return self.schedule_items.filter(slug=slug).first()
 
     @strawberry.field
-    def ranking(self, info) -> Optional[RankRequest]:
+    def ranking(self, info, topic: str) -> Optional[RankRequest]:
```
marcoacierno, line 313:
```suggestion
    def ranking(self, info, topic: ID) -> Optional[RankRequest]:
```

frontend/src/pages/ranking/index.tsx
```diff
@@ -150,11 +166,18 @@ export const RankingPage: React.FC = () => {
 
 export const getStaticProps: GetStaticProps = async ({ locale }) => {
   const client = getApolloClient();
-
+  console.log("GetStaticProps ranking");
```
marcoacierno, line 169:
```suggestion
```

frontend/src/pages/ranking/index.tsx
```diff
@@ -150,11 +166,18 @@ export const RankingPage: React.FC = () => {
...
+    data: {
+      conference: { topics },
+    },
+  } = await prefetchTopics(client, locale);
 
   return addApolloState(client, {
-    props: {},
+    props: {
+      topics: topics,
+    },
```
marcoacierno, line 180: you don't need all of this

We use Apollo cache to get the data, so you only need to call `queryTopics`

estyxx: But why re-execute the query (ok the second time will be cached...) instead of just give the topics to the component with `props`?

frontend/src/pages/ranking/index.tsx
```diff
@@ -48,6 +58,15 @@ export const RankingPage: React.FC = () => {
     return true;
   };
 
+  const [filters, { select }] = useFormState<Filters>(
+    {},
+    {
+      onChange(e, stateValues, nextStateValues) {
+        setCurrentTopic(nextStateValues.topic);
+      },
```
marcoacierno, line 66: uhm why the state? can't you use filters.values?

frontend/src/pages/ranking/index.tsx
```diff
@@ -150,11 +166,18 @@ export const RankingPage: React.FC = () => {
 
 export const getStaticProps: GetStaticProps = async ({ locale }) => {
   const client = getApolloClient();
-
+  console.log("GetStaticProps ranking");
   await prefetchSharedQueries(client, locale);
```
marcoacierno, line 170:
```suggestion
  await Promise.all([prefetchSharedQueries(client, locale), queryTopics(client, {
    code: process.env.conferenceCode,
  })]);
```

frontend/src/pages/ranking/index.tsx
marcoacierno: you don't refetch, the first time it will come from the cache

then you will be able to reuse the same query when the topic changes

backend/api/conferences/types.py
```diff
@@ -323,11 +323,12 @@ def ranking(self, info) -> Optional[RankRequest]:
             ):
                 return None
 
+            submissions = rank_request.rank_submissions.filter(
+                submission__topic__id=int(topic)
```
marcoacierno, line 327:
```suggestion
                submission__topic__id=topic
```

I don't think we need to cast here, django should handle it 👍

frontend/src/pages/ranking/index.tsx
```diff
@@ -29,15 +29,31 @@ const COLORS = [
...
+  const {
+    data: {
+      conference: { topics },
+    },
+  } = useTopicsQuery({
+    variables: {
+      code: conferenceCode,
+    },
+  });
+  const [filters, { select }] = useFormState<Filters>({ topic: topics[0].id });
```
marcoacierno, line 48: What is the sorting? database dependant? or are topics ordered?

estyxx: Think database dependent, they arrive order by Id

marcoacierno: uhm oki, well that's ok for now I guess :)

frontend/src/helpers/prefetch.ts
```diff
@@ -1,6 +1,6 @@
 import { ApolloClient } from "@apollo/client/core";
 
-import { queryFooter, queryHeader } from "~/types";
+import { queryFooter, queryHeader, queryTopics } from "~/types";
```
marcoacierno, line 3:
```suggestion
import { queryFooter, queryHeader } from "~/types";
```

frontend/src/pages/ranking/index.tsx
```diff
@@ -1,7 +1,7 @@
 /** @jsxRuntime classic */
 
 /** @jsx jsx */
-import React from "react";
+import React, { useState } from "react";
```
marcoacierno, line 4:
```suggestion
import React from "react";
```

marcoacierno (approved):

frontend/src/pages/ranking/index.tsx
```diff
@@ -93,10 +109,7 @@ export const RankingPage: React.FC = () => {
                   borderRadius: 0,
                 }}
               >
-                <FormattedMessage id="voting.allTopics">
-                  {(text) => <option value="">{text}</option>}
-                </FormattedMessage>
-                {data?.conference?.topics.map((topic) => (
```
marcoacierno, line 99: we fetched the topics here?

PR #2731, opened 2022-03-29
Implement generating speakers voucher codes

marcoacierno:
## How to test it

image


backend/conferences/admin.py
```diff
@@ -102,6 +102,10 @@ class ConferenceAdmin(admin.ModelAdmin):
             "Pretix",
             {"fields": ("pretix_organizer_id", "pretix_event_id", "pretix_event_url")},
         ),
+        (
+            "Speaker Voucher",
+            {"fields": ("pretix_speaker_voucher_quota_id",)},
```
estyxx, line 107: Just curious: why another pretix id?

backend/schedule/admin.py
```diff
@@ -25,9 +28,85 @@
     ScheduleItemAttendee,
     ScheduleItemInvitation,
     Slot,
+    SpeakerVoucher,
 )
 
 
+@admin.action(description="Generate voucher codes")
+def generate_voucher_codes(modeladmin, request, queryset):
```
estyxx, line 36: We need to refactor the admins, they have a lot of logic... we should move in some service level this stuff...

estyxx (approved):

backend/conferences/admin.py
marcoacierno: this is the ID of the quota to use to assign to the vouchers code

PR #2777, opened 2022-04-18
Use Pretix API to check for attendee's ticket

marcoacierno:
## Why

We are getting rid of direct DB calls to integrate with Pretix to allow us to test APIs locally and on staging (both do not have access to the pretix DB)

We implemented an API on pretix to check if a person has a ticket, this PR now remove the DB equivalent and uses that API

### DIFF
--- a/backend/pretix/__init__.py
+++ b/backend/pretix/__init__.py
@@ -1,5 +1,6 @@
 import logging
 import typing
+from dataclasses import dataclass
 from datetime import date
 from urllib.parse import urljoin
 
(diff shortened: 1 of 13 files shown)


backend/pretix/__init__.py
```diff
@@ -347,3 +348,34 @@ def create_order(conference: Conference, order_data: CreateOrderInput) -> Order:
     data = response.json()
 
     return Order(code=data["code"], payment_url=data["payments"][0]["payment_url"])
+
```
estyxx, line 351: can we move this logic out of `__init__.py` file sooner or later? As I don’t expect to find anything inside usually and it’s the last place I’d look...

estyxx (approved):

backend/pretix/__init__.py
marcoacierno: in the future the pretix API will live in pythonit-toolkit

backend/pretix/__init__.py
```diff
@@ -347,3 +348,34 @@ def create_order(conference: Conference, order_data: CreateOrderInput) -> Order:
...
+
+@dataclass
+class Conference:
+    pretix_organizer_id: str
+    pretix_event_id: str
+
+
+def user_has_admission_ticket(
+    *, email: str, event_organizer: str, event_slug: int
```
estyxx, line 360:
```suggestion
    *, email: str, event_organizer: str, event_slug: str
```
Think the `event_slug` is a string as well

PR #2788, opened 2022-04-25
Implement users tickets

estyxx:
## Why

Add tickets to User type. 
Moved the construction of the Strawberry types inside the class itself in order to have fewer little scattered methods and be able to reuse some of the types for User's tickets.


backend/api/pretix/types.py
```diff
@@ -1,10 +1,19 @@
...
 import strawberry
 
 from api.pretix.constants import ASSOCIATION_CATEGORY_INTERNAL_NAME
+from pretix.types import Category as CategoryDict
+from pretix.types import Item as ItemDict
+from pretix.types import Option as OptionDict
+from pretix.types import OrderPosition as OrderPositionDict
+from pretix.types import ProductVariation as ProductVariationDict
+from pretix.types import Question as QuestionDict
+from pretix.types import Quota as QuotaDict
```
marcoacierno, line 16: weird it didn't get formatted in one import

backend/api/pretix/types.py
```diff
@@ -92,6 +163,86 @@ def type(self) -> TicketType:
...
+            ],
+            questions=[
+                Question.from_data(question, language)
+                for question in questions
+                if data["id"] in question["items"]
+            ],
+            quantity_left=cls._get_quantity_left(data, quotas),
+        )
+
+    @staticmethod
```
marcoacierno, line 200: why a staticmethod?

backend/api/pretix/types.py
```diff
@@ -92,6 +163,86 @@ def type(self) -> TicketType:
...
+        # https://docs.pretix.eu/en/latest/development/concepts.html#quotas
+        return min(
+            quota["available_number"]
+            for quota in quotas.values()
+            if int(data["id"]) in quota["items"]
+        )
+
+
+@strawberry.type
+class PretixOrderPosition:
```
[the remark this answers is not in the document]
estyxx: oh I like `AttendeeTicket`! I didn't know how to call it and just `Ticket` I thought was a bit confusing because there is already the `PretixTicket` above...

backend/api/pretix/types.py
estyxx: because I'm using it in `from_data` that is a `classmethod`, so I don't have the object instance... can be static or class method

backend/api/pretix/types.py
estyxx: There is a specific configuration for `as` imports: `combine_as_imports` by default is False!

marcoacierno (changes requested):
I think once you rename the type this PR should be good to go

backend/pretix/__init__.py
```diff
@@ -347,3 +344,18 @@ def create_order(conference: Conference, order_data: CreateOrderInput) -> Order:
...
     return Order(code=data["code"], payment_url=data["payments"][0]["payment_url"])
+
+
+def get_user_tickets(conference: Conference, email: str):
+    response = pretix(
+        conference=conference,
+        endpoint="tickets/attendee-tickets",
+        method="get",
+        qs={
+            "attendee_email": email,
```
marcoacierno, line 355: does the query string appear in the logs? I am worried it might cause GDPR issues

marcoacierno: cloudwatch logs get deleted every 2 weeks so it might be fine for now!

backend/pretix/__init__.py
```diff
@@ -347,3 +344,18 @@ def create_order(conference: Conference, order_data: CreateOrderInput) -> Order:
     data = response.json()
 
     return Order(code=data["code"], payment_url=data["payments"][0]["payment_url"])
+
+
+def get_user_tickets(conference: Conference, email: str):
+    response = pretix(
+        conference=conference,
+        endpoint="tickets/attendee-tickets",
```
marcoacierno, line 352: maybe sure you don't need to add a forward slash here

marcoacierno (approved):
tried buying a ticket, everything worked fine so feel free to merge 👍

PR #2802, opened 2022-05-02
Add text-friendly schedule version

marcoacierno:
## Why

image

image


estyxx (approved):
Would be nice to have a button in the schedule as well. So if anybody else needs it's there... 

Mmmm would be nice as well to know how many people actually printed this... 🤔

PR #2812, opened 2022-05-05
Add "My Tickets" section in the profile

estyxx:
- WIP Implement user tickets api
- Refactor types construction
- Add type annotations
- Revert test_mailchimp
- Add type annotations
- Move api tests
- Combine as imports
- Rename PretixOrderPosition to AttendeeTicket
- WIP
- WIP tickets
- WIP Update AttendeeTicket Mutation


frontend/src/components/modal/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <Box
+          sx={{
+            position: "fixed",
+            inset: 0,
+            transition: "opacity 150ms cubic-bezier(0.4, 0, 0.2, 1)",
+            bg: "gray",
+            opacity: 0.75,
+          }}
+          onClick={onClose}
+        ></Box>
```
marcoacierno, line 46:
```suggestion
        />
```

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,65 @@
...
+from api.pretix.query import get_user_tickets
+from api.pretix.types import AttendeeTicket, UpdateAttendeeTicketInput
+from conferences.models.conference import Conference
+
+logger = logging.getLogger(__name__)
+
+
+@strawberry.type
+class UpdateAttendeeTicketError:
+    message: str = ""
```
marcoacierno, line 17: why this empty string?

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,65 @@
...
+@strawberry.type
+class AttendeeTicketMutation:
+    @strawberry.mutation(permission_classes=[IsAuthenticated])
+    def update_attendee_ticket(
+        self, info: Info, conference: str, input: UpdateAttendeeTicketInput
+    ) -> UpdateAttendeeTicketResult:
+        conference = Conference.objects.get(code=conference)
+
+        try:
+            pretix.update_ticket(conference, input)
```
marcoacierno, line 41: make sure you add a permission check so people can't edit other people tickets

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,65 @@
...
+
+            # TODO: filter by orderposition
+            tickets = get_user_tickets(
+                conference, info.context.request.user.email, language="en"
+            )
+
+            tickets = list(filter(lambda ticket: str(ticket.id) == input.id, tickets))
+            if tickets:
+                return tickets[0]
+
```
marcoacierno, line 51: I don't understand why we need this, doesn't Pretix return you the updated order position in the original update_ticket call?

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,65 @@
...
+        except Exception as e:
```
marcoacierno, line 52: I don't like this generic exception

I would feel more comfortable with a specific one, what error do you want to catch?

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,65 @@
...
+                e,
+                exc_info=True,
+            )
+            return UpdateAttendeeTicketError(e)
+
+        # If the user has changed the email, the ticket will not be returned but
+        # the mutation succeeded.
+        return TicketReassigned(
+            id=input.id, message=TicketReassigned.message.format(email=input.email)
+        )
```
marcoacierno, line 65: Why a message? I don't believe the BE should return messages, that's the FE's responsibility. You should return the new assigned email as a field, and even then you might not need it since the FE already has the new email

backend/pretix/__init__.py
```diff
@@ -30,6 +31,7 @@ def pretix(
     method="get",
     **kwargs,
 ):
+    print(get_api_url(conference, endpoint))
```
marcoacierno, line 34:
```suggestion
```

backend/pretix/__init__.py
```diff
@@ -384,3 +386,18 @@ def get_user_tickets(conference: Conference, email: str):
     )
 
     return response.json()
+
+
+def update_ticket(conference: Conference, attendee_ticket: UpdateAttendeeTicketInput):
+    print("Updating Ticket")
```
marcoacierno, line 392: make sure you remove all those prints once done

to debug you can also use `breakpoint()` that might be easier than having to look at the console

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { AttendeeTicket, useUpdateTicketMutation } from "~/types";
+
+import { ProductState } from "../tickets-page/types";
+import { Link } from "~/components/link";
+
+type Props = {
+  tickets?: AttendeeTicket[];
+};
+
+export const MyTickets: React.FC<Props> = ({ tickets }) => {
```
marcoacierno, line 25: you don't need React.FC if you don't use `children`



```suggestion
export const MyTickets = ({ tickets }: Props) => {
```

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { Table } from "~/components/table";
+import { QuestionsSection } from "~/components/tickets-page/questions-section";
+import { useTranslatedMessage } from "~/helpers/use-translated-message";
+import { AttendeeTicket, useUpdateTicketMutation } from "~/types";
+
+import { ProductState } from "../tickets-page/types";
+import { Link } from "~/components/link";
+
+type Props = {
+  tickets?: AttendeeTicket[];
```
marcoacierno, line 22: why the `?` can it really be undefined? I would prefer an empty list 🤔

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+};
+
+export const MyTickets: React.FC<Props> = ({ tickets }) => {
+  const code = process.env.conferenceCode;
+  const ticketHeader = useTranslatedMessage("profile.ticketFor");
+  const nameHeader = useTranslatedMessage("orderReview.attendeeName");
+  const emailHeader = useTranslatedMessage("orderReview.attendeeEmail");
+  const [currentTicket, setCurrentTicket] = useState({ id: null, show: false });
```
marcoacierno, line 30: why do you need a dict?

if `id` is null => don't show anything
if `id` is not null => show the modal

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+  const headers = [ticketHeader, nameHeader, emailHeader, ""];
+
+  const [selectedProducts, setSelectedProducts] = useState(
```
marcoacierno, line 34: why `selectedProducts`? is this the list of tickets?

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+          px: 3,
+        }}
+      >
+        <Heading mb={5} as="h2" sx={{ fontSize: 5 }}>
+          <FormattedMessage id="profile.myTickets" />
+        </Heading>
+        {tickets.length === 0 && (
+          <FormattedMessage
+            id="profile.myTickets.notickets"
+            values={{ linkTicket: <Link path="/tickets">here</Link> }}
```
marcoacierno, line 147: `here` is not translated in italian

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+          <Alert variant="success">
+            <FormattedMessage
+              id={`profile.myTickets.update.AttendeeTicket.message`}
+            />
+          </Alert>
+        )}
+        {updatedData?.updateAttendeeTicket.__typename ===
+          "TicketReassigned" && (
+          <Alert variant="success">
+            {updatedData.updateAttendeeTicket.message}
```
marcoacierno, line 182: as I said above, I would do this:

```suggestion
            <FormattedMessage id="profile.tickets.ticketReassigned" defaultMessage="The ticket has been re-assigned to {email}!" />
```

so you can translate the message in italian

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+              </Box>,
+            ]}
+          />
+        )}
+
+        {updatedData?.updateAttendeeTicket.__typename === "AttendeeTicket" && (
+          <Alert variant="success">
+            <FormattedMessage
+              id={`profile.myTickets.update.AttendeeTicket.message`}
+            />
```
marcoacierno, line 176:
```suggestion
            <FormattedMessage
              defaultMessage="English Message here"
              id="profile.myTickets.update.AttendeeTicket.message"
            />
```

I am trying to have the default english message in the code, what do you think?

backend/api/pretix/mutations.py
estyxx: I can add a default message. 
I would like to render the message based on the selected language so on the frontend I just have a generic locale message set when something went wrong

backend/api/pretix/mutations.py
estyxx: What if the pretix api is down?

backend/api/pretix/mutations.py
estyxx: Yes, but not will all the information we need: the item it's just an ID instead of the dictionary, the question as well... 
in order to return AttendeeTicket type I need that information... 
At least the questions...

backend/api/pretix/mutations.py
marcoacierno: it is not just about pretix, with `except Exception` you are catching a lot more, even code errors like:

```
>>> try:
...     5/0
... except Exception:
...     print("hello")
... 
hello
```

marcoacierno: I am ok with the API returning 500 or unhandled errors in graphql if something un expected happens

frontend/src/locale/index.ts
```diff
@@ -46,8 +46,14 @@ export const messages = {
     "profile.status": "Status",
     "profile.price": "Price",
     "profile.manageOrder": "Manage order",
-    "profile.items": "Tickets",
+    "profile.myTickets": "My Tickets",
+    "profile.myTickets.notickets":
+      "No tickets assigned to your email, you can buy one {linkTicket}",
```
marcoacierno, line 51: another thing that can help here is to show the email and say something like this:

```suggestion
    "profile.myTickets.notickets":
      "No tickets found assigned to your email: {email}. Make sure the attendee email and your email matches in your order or buy a ticket {linkTicket}",
```

so users know they need to update the attendee email to match the tickets

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,75 @@
...
+    @strawberry.mutation(permission_classes=[IsAuthenticated])
+    def update_attendee_ticket(
+        self, info: Info, code: str, input: UpdateAttendeeTicketInput
+    ) -> UpdateAttendeeTicketResult:
+        conference = Conference.objects.get(code=code)
+        try:
+            email = info.context.request.user.email
+            if not is_ticket_owner(conference, email, input.id):
+                return UpdateAttendeeTicketError(
+                    message="You are not allowed to updtae this ticket."
```
marcoacierno, line 48:
```suggestion
                    message="You are not allowed to update this ticket."
```

marcoacierno: we could consider the idea of just saying "This ticket does not exist" for security purposes, but it is not super important

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,228 @@
...
+  const ticketHeader = useTranslatedMessage("profile.ticketFor");
+  const nameHeader = useTranslatedMessage("orderReview.attendeeName");
+  const emailHeader = useTranslatedMessage("orderReview.attendeeEmail");
+  const [currentTicket, setCurrentTicket] = useState({ id: null, show: false });
+
+  const headers = [ticketHeader, nameHeader, emailHeader, ""];
+
+  const [selectedProducts, setSelectedProducts] = useState(
+    tickets.reduce((acc, ticket) => {
+      (acc[ticket.id] = acc[ticket.id] || []).push({
```
marcoacierno, line 36: I don't understand this code `(acc[ticket.id] = acc[ticket.id] || [])` 🤔

estyxx: So the first time `accumulator[ticket.id]` is undefined, so this or means assign to `accumulator[ticket.id]` `accumulator[ticket.id]` if there is a value otherwise assign `[]`

frontend/src/components/profile/my-tickets.tsx
estyxx: `selectedProducts` is needed to `QuestionSection` to render correctly

backend/api/pretix/mutations.py
estyxx: I add a `IsOwnerPermission`

frontend/src/components/profile/my-tickets.tsx
estyxx: Mmmm I find it unnecessary
1. we will not notice if we are missing a translation because there is the default message
2. If we have to change the message we have to change in two places
3. Some messages are quite long...

marcoacierno: >If we have to change the message we have to change in two places

we already have to change it in two places

> we will not notice if we are missing a translation because there is the default message

not sure of this, we already missed this quite a lot of times

> Some messages are quite long...

not sure how much it matters

one pro side of having the message as `defaultMessage` is that you will be able to find the code much easier when searching the string instead of having to find the locale.ts file -> copying the key -> search again

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,48 @@
...
+    "UpdateAttendeeTicketResult",
+    (TicketReassigned, AttendeeTicket),
+)
+
+
+@strawberry.type
+class AttendeeTicketMutation:
+    @strawberry.mutation(permission_classes=[IsTicketOwner])
+    def update_attendee_ticket(
+        self, info: Info, conference_code: str, input: UpdateAttendeeTicketInput
```
marcoacierno, line 31:
```suggestion
        self, info: Info, conference_code: str, input: UpdateAttendeeTicketInput
```

is this `conference_code` consistent with other APIs?

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,48 @@
...
+    ) -> UpdateAttendeeTicketResult:
+        conference = Conference.objects.get(code=conference_code)
+
+        pretix.update_ticket(conference, input)
+
+        # TODO: filter by orderposition
+        tickets = get_user_tickets(
+            conference, info.context.request.user.email, language="en"
```
marcoacierno, line 39: why the hardcoded language? can't you add the locale as input?

marcoacierno: what happens if I edit the ticket and I am in the Italian locale, after update the ticket name will become engilsh?

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,48 @@
...
+        )
+
+        tickets = list(filter(lambda ticket: str(ticket.id) == input.id, tickets))
+        if tickets:
+            return tickets[0]
```
marcoacierno, line 44:
```suggestion
        ticket = next(filter(lambda ticket: str(ticket.id) == input.id, tickets), None)
        if ticket:
            return ticket
```

maybe?

backend/api/pretix/permissions.py
```diff
@@ -0,0 +1,24 @@
...
+class IsTicketOwner(BasePermission):
+    message = "You are not allowed to update this ticket."
+
+    def has_permission(self, source, info, **kwargs):
+        pastaporto = info.context.request.pastaporto
+
+        if not pastaporto.is_authenticated:
+            return False
+
+        conference = Conference.objects.get(code=kwargs["conferenceCode"])
```
marcoacierno, line 16: why is the conference code not in snake case here 🤔

backend/api/pretix/tests/test_update_attendee_ticket.py
```diff
@@ -0,0 +1,228 @@
...
+                "email": "ester@pycon.it",
+            },
+        },
+    )
+
+    assert "404 Client Error" in response["errors"][0]["message"]
+
+
+@override_settings(PRETIX_API="https://pretix/api/")
+def test_update_ticket(
```
marcoacierno, line 84: What is the original name that you change in this test?

backend/pretix/__init__.py
```diff
@@ -383,4 +384,38 @@ def get_user_tickets(conference: Conference, email: str):
...
     )
 
+    response.raise_for_status()
+
+    return response.json()
+
+
+def get_user_ticket(conference: Conference, email: str, id: str):
+
+    # TODO: filter by orderposition in the PretixAPI
```
marcoacierno, line 394: what is this todo? what is missing?

backend/pretix/__init__.py
```diff
@@ -383,4 +384,38 @@ def get_user_tickets(conference: Conference, email: str):
...
+        return str(ticket["id"]) == id and ticket["attendee_email"] == email
+
+    tickets = list(filter(filter_by, tickets))
+
+    if tickets:
+        return tickets[0]
+
+
+def is_ticket_owner(conference: Conference, email: str, id: str) -> bool:
+    ticket = get_user_ticket(conference, email, id)
```
marcoacierno, line 407: we should add a new API in pretix that checks the ownership

this API fetches lots more information than we need

but this is for after we merge this

frontend/src/components/profile/my-tickets.tsx
marcoacierno: I feel like we failed to make the form generic

but doesn't make sense to work on it now since it is late, we will refactor the tickets page in the future anyway

backend/api/pretix/mutations.py
estyxx: Think only [order](https://github.com/pythonitalia/pycon/blob/main/backend/api/orders/query.py#L13) query. 
I will switch back to `conference`...

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,264 @@
...
+          },
+          variables: {
+            conference: process.env.conferenceCode,
+            language: language,
+          },
+        });
+      }
+    },
+    onError(err) {
+      console.error(err);
```
marcoacierno, line 103: do we need this console.error?

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,264 @@
...
+      setError(err.message);
```
marcoacierno, line 104: why do we have this setError? isn't it the same of `updatedError`?

frontend/src/components/profile/my-tickets.tsx
```diff
@@ -0,0 +1,264 @@
...
+    },
+  });
+
+  const updateTicketCallback = useCallback(
+    (id: string) => {
+      return () => {
+        const answers = tickets
+          .filter((item) => item.id == id)[0]
```
marcoacierno, line 112: all of those hard coded `[0]` around this code make me feel a bit uneasy 😓

backend/api/pretix/permissions.py
estyxx: I didn't investigate, seems strawberry is passing the bare `input` dictionary to the `has_permision`...

backend/api/pretix/tests/test_update_attendee_ticket.py
estyxx: PRETIX_API: https://tickets.pycon.it/api/v1/

backend/pretix/__init__.py
estyxx: Would be nice to add this thing to the API directly so I don't have to fetch all the tickets to get only one...

backend/pretix/__init__.py
estyxx: agree

frontend/src/components/profile/my-tickets.tsx
estyxx: yeah, in order to make it work with `QuestionsSections` I have these weird structures... Also, because I render only one ticket at a time, not all the tickets like on the `/tickets` page...

backend/api/pretix/types.py
```diff
@@ -165,6 +167,12 @@ def type(self) -> TicketType:
 
         return TicketType.STANDARD
 
+    @strawberry.field
+    def sold_out(self) -> Optional[bool]:
```
estyxx, line 171: I removed the custom `Ticket` type in the frontend to use those generated by the backend, so soldOut was calculated on the frontend. I left the calculation in the frontend but I added this field so typescript will not complain and later we can use this instead...

frontend/src/components/profile/my-tickets.tsx
estyxx: yeah, seems quite complicated... 😄

frontend/src/components/profile/my-tickets.tsx
estyxx: On the `/tickets` page we group by the `TicketItem.id` so we can have many items for one id, here I grouped by the `AttendeeTicket.id` that is unique, so there will be only one item.

backend/api/pretix/types.py
marcoacierno: What problem did you have?

marcoacierno:
we should disable the scroll when the modal is open

maybe we already did it somewhere else and you can copy it?

marcoacierno:
image

save fails saying questions are required even when they are not:

image

backend/api/pretix/types.py
estyxx: The build was failing in the typescript checks because I was passing `TicketItem` down to `QuestionSection`, we didn't have all these types in the backend before I just created them... so Typescript was complaining `Ticket` is not compatible with `TicketItem` all over... 
or stuff like `soldOut` is not in `TicketItem`.... etc etc

estyxx:
> image
> 
> save fails saying questions are required even when they are not:
> 
> image

Mmmm interesting, Pretix is complaining anyway... 
```
{'answers': [{}, {'answer': ['This field may not be blank.']}, {}, {}]}
```
This is the reply I got from pretix...

estyxx:
Yes if you don't include an answer in Pretix's API call, it will delete the answer... So it should not complain for required empty answers @marcoacierno  now :)

marcoacierno:
> we should disable the scroll when the modal is open
> 
> maybe we already did it somewhere else and you can copy it?

can you look into disabling the scroll too when the modal is open

backend/api/pretix/mutations.py
```diff
@@ -0,0 +1,95 @@
...
+
+def _get_update_tickets_errors(
+    response, input: UpdateAttendeeTicketInput
+) -> UpdateAttendeeTicketErrors:
+    errors = []
+    for field in ("attendee_name", "attendee_email"):
+        if response.get(field):
+            errors.append(
+                UpdateAttendeeTicketError(field=field, message=response[field][0])
+            )
```
marcoacierno, line 81: I tried sending the mutation without attendee email and it saved it empty

the email should always be required

estyxx: Fixed. Weird that pretix allows it!!

estyxx:
> > we should disable the scroll when the modal is open
> > maybe we already did it somewhere else and you can copy it?
> 
> can you look into disabling the scroll too when the modal is open?

Yeah... I'm trying to set `overflow: "hidden"` in the Modal but seems not working... mmm

backend/api/pretix/tests/test_get_conference_tickets.py
```diff
@@ -75,7 +75,7 @@ def test_get_conference_tickets(
     assert ticket.available_until is None
     assert ticket.quantity_left == 118
     assert ticket.questions[0].id == 1
-    assert ticket.questions[0].name == "Codice Fiscale"
```
marcoacierno, line 78: what happened here?

marcoacierno (approved):

PR #2818, opened 2022-05-07
Badge page

marcoacierno:


frontend/src/pages/badge.tsx
```diff
@@ -0,0 +1,166 @@
...
+      >
+        <Logo
+          sx={{
+            width: "100px",
+          }}
+        />
+
+        <Text
+          sx={{
+            fontSize: "36px",
```
patrick91, line 112: maybe we should use `pt` for print?

marcoacierno: I used this to convert everything to points: https://simplecss.eu/pxtoems.html

PR #3123, opened 2022-11-20
Link grant with the logged user

estyxx:


backend/api/grants/forms.py
```diff
@@ -23,13 +23,23 @@ def clean(self):
         if not conference.is_grants_open:
             raise exceptions.ValidationError(_("The grants form is now closed!"))
 
+        try:
+            Grant.objects.get(user_id=self.context.request.user.id)
+            raise exceptions.ValidationError(_("Grant already submitted!"))
+        except Grant.DoesNotExist:
+            pass
+
```
marcoacierno, line 31:
```
if Grant.objects.filter(user_id=self.context.request.user.id).exists()
```

this way django doesn't do an a select query fetching data/creating the object just to throw it away

frontend/src/pages/grants/index.tsx
```diff
@@ -87,13 +92,31 @@ export const GrantsPage = () => {
           my: 5,
         }}
       >
-        {status === DeadlineStatus.HappeningNow && (
-          <GrantForm conference={code} />
-        )}
-        {status === DeadlineStatus.InTheFuture && (
-          <GrantsComingSoon start={start} />
+        {isLoggedIn ? (
```
marcoacierno, line 95: what happens if we remove this condition?

estyxx: the page already redirect! I forgot to remove this!

marcoacierno (approved):

PR #3136, opened 2022-11-28
Grants: replace Age field with Group Age

estyxx:


backend/grants/admin.py
```diff
@@ -13,7 +13,6 @@
     "name",
     "full_name",
     "email",
-    "age",
```
marcoacierno, line 16: did you forget to add age_group here?

frontend/src/components/grant-form/index.tsx
```diff
@@ -66,12 +65,12 @@ export const GrantForm = ({ conference }: Props) => {
       formState.setField("fullName", user.fullName);
       formState.setField("name", user.name);
       formState.setField("gender", user.gender);
-      if (user.dateBirth) {
-        formState.setField(
-          "age",
-          new Date().getFullYear() - new Date(user.dateBirth).getFullYear(),
-        );
-      }
+      // if (user.dateBirth) {
```
marcoacierno, line 68: do we want to automatically pre-select the age group?

frontend/src/locale/index.ts
```diff
@@ -354,7 +354,18 @@ export const messages = {
     "grants.form.fields.email": "Email",
     "grants.form.fields.email.description":
       "Here you will receive a notification about your application.",
-    "grants.form.fields.age": "Age",
+    "grants.form.fields.ageGroup": "Age",
+    "grants.form.fields.ageGroup.values.selectOption": "Select Option",
+    "grants.form.fields.ageGroup.values.range_less_than_10":
```
marcoacierno, line 359: is anyone ever going to select this :P 😂

backend/grants/models.py
```diff
@@ -25,6 +25,16 @@
 
 
 class Grant(TimeStampedModel):
+    class AgeGroup(models.TextChoices):
+        RANGE_LESS_THAN_10 = "range_less_than_10", _("10 years old or under")
```
marcoacierno, line 29: I am not 100% convinced about having this "less than 10 years old" option, it is not like someone is going to use it 🤔 

I need to think about it a bit

frontend/src/locale/index.ts
estyxx: Yes, at PyCon 11 we had one grant request for a 9 years old.

frontend/src/components/grant-form/index.tsx
estyxx: yes I was thinking about how to do it in an elegant way... :D

backend/grants/admin.py
estyxx: No, I will look at the export in another PR, as I need to remove some stuff now that we have the user in the model.

marcoacierno (approved):

PR #3197, opened 2022-12-28
Update Ranking Algorithm

estyxx:


backend/voting/models/ranking.py
```diff
@@ -35,9 +36,15 @@ def __str__(self):
...
+class Rank(TypedDict):
+    submission_id: int
+    submission__tag: SubmissionTag
+    score: float
+
+
 class RankRequest(models.Model):
 
-    conference = models.OneToOneField(
+    conference = models.ForeignKey(
```
estyxx, line 47: I relaxed this constraint to be able to see both the old ranking and the new one with the tags in order to check them, but I think I will revert this, we don't need to have multiple rankings for one conference.

marcoacierno (approved):

PR #3266, opened 2023-01-30
Grants export add admin link

estyxx:

### DIFF
--- a/backend/grants/admin.py
+++ b/backend/grants/admin.py
@@ -38,6 +38,7 @@ class GrantResource(ResourceUsersByIdsMixin):
     submission_tags = Field()
     submission_admin_link = Field()
     submission_pycon_link = Field()
+    grant_admin_link = Field()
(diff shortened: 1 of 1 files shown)


backend/grants/admin.py
```diff
@@ -95,6 +96,9 @@ def dehydrate_submission_admin_link(self, obj):
             ]
         )
 
+    def dehydrate_grant_admin_link(self, obj: Grant):
+        return f"https://admin.pycon.it/admin/grants/grant/?q={'+'.join(obj.full_name.split(' '))}"
```
marcoacierno, line 100: weird link 😂 why not the detail page?

marcoacierno (approved):

backend/grants/admin.py
estyxx: because in the change page you will not see the email clearly.

estyxx: probably wee need to discuss how to send them, they are too many :D

PR #3299, opened 2023-02-10
Grants: fix Plain

estyxx:
- Use email as Customer identifier
- Add Sentry to aws lambda

### DIFF
--- a/backend/wsgi_handler.py
+++ b/backend/wsgi_handler.py
@@ -14,14 +14,32 @@ import sys
 import traceback
 from logging import getLogger
 
+import sentry_sdk
 from django.apps import apps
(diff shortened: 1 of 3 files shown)


backend/wsgi_handler.py
```diff
@@ -24,6 +27,21 @@
 logger = getLogger(__name__)
 
 
+SENTRY_DSN = settings.SENTRY_DSN
+
+if SENTRY_DSN:
```
marcoacierno, line 32: is this present somewhere else too?

estyxx: settings.prod

marcoacierno (approved):

PR #3300, opened 2023-02-10
Grants Admin: add recap by Country

estyxx:


backend/countries/countries.py
```diff
@@ -0,0 +1,91 @@
...
+
+from dataclasses import dataclass
+from typing import Any, Dict, Generic, List, Optional, TypeVar, Union
+
+from countries.constants import CONTINENTS, COUNTRIES
+
+T = TypeVar("T", bound="Base")
+
+
+class Base(Generic[T]):
```
marcoacierno, line 11: I don't think I understand this code, can you explain what it is?

backend/grants/admin.py
```diff
@@ -411,3 +417,76 @@ def save_form(self, request, form, change):
...
+                    "waiting_for_confirmation": counter.get(
+                        Grant.Status.waiting_for_confirmation, 0
+                    ),
+                    "refused": counter.get(Grant.Status.refused, 0),
+                    "confirmed": counter.get(Grant.Status.confirmed, 0),
+                }
+            )
+
+        results = sorted(results, key=lambda k: k["count"], reverse=True)
+        results = sorted(results, key=lambda k: (k["continent"],))
```
marcoacierno, line 472: wasn't it possible to do it in one call to `sorted`?

 ```suggestion
        results = sorted(results, key=lambda k: (-k["count"], k["continent"]))
```

something like this? I might be wrong, I didn't try it

File: backend/grants/models.py [hunk 3]
```diff
@@ -110,7 +110,6 @@ class ApprovedType(models.TextChoices):
     age_group = models.CharField(
         _("Age group"), max_length=20, choices=AgeGroup.choices, blank=True
     )
-    age = models.PositiveSmallIntegerField(_("age"), null=True)  # TODO: remove
```
marcoacierno, line 113: do we want to remove this field? I guess we won't need the legacy data

backend/users/models.py
```diff
@@ -81,18 +33,3 @@ def __str__(self):
...
-
-    def is_italian(self):
-        if self.country == "IT":
-            return True
-        return False
-
-    @property
-    def submissions(self):
-        # TODO: Temp method to make the migration easier
-        return Submission.objects.filter(speaker_id=self.id)
```
marcoacierno, line 98: do we not need this submissions property anymore?

backend/countries/countries.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def filter(self, **kwargs):
+        return iter(
+            c
+            for c in self.items
+            if any(getattr(c, k, None) == v for k, v in kwargs.items())
+        )
+
+
+countries = Countries()
```
marcoacierno, line 91: I feel like this whole file could be simpler, since the only thing we do is `countries.get`? what do you think?

backend/grants/models.py
estyxx: yes, I want to remove this since now we have AgeGroup

backend/grants/models.py
```diff
@@ -147,6 +153,19 @@ class ApprovedType(models.TextChoices):
     def __str__(self):
         return f"{self.full_name}"
 
+    def save(self, *args, **kwargs):
+        if self.traveling_from:
+            country = countries.get(code=self.traveling_from)
+            assert country
```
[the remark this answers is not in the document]
estyxx: think was because of typing: get can potentially return None, very unlikely... I can put an if just in case

backend/countries/countries.py
estyxx: I want the list of countries/continents to be "smarter" than a bare list of dictionaries. Since both Countries and Continents use the same logic I wrote a base class and use Generics for the typing...

backend/countries/countries.py
estyxx: In the end I grouped the grants only by Country and didn't actually use the Continent info. It was quite difficult to override the change_list view and put the data on it. Ideally, I would like to group them by Continent and Country, in this case, this division can be useful

backend/countries/countries.py
estyxx: I want to mimic the behaviour of the PyCountries library and have the same api

backend/countries/countries.py
estyxx: probably `filter` is not needed

backend/grants/models.py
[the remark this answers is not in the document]
marcoacierno: I guess we can just not have any if and have a null error if anything weird happens?

marcoacierno (approved):
feel free to merge if you like the code and everything works 👍

backend/countries/countries.py
marcoacierno: I am not a big fan of this code, we could just implement the API/interface that we need or a simpler one since our code is not a library, but up to you

PR #3301, opened 2023-02-11
Grants Admin: remove plain action, use fieldset 💅

estyxx:


backend/grants/admin.py
```diff
@@ -317,12 +310,59 @@ class GrantAdmin(ExportMixin, AdminUsersMixin, SearchUsersMixin):
     )
     user_fk = "user_id"
     actions = [
-        send_messages_to_plain,
         send_reply_emails,
         send_grant_reminder_to_waiting_for_confirmation,
         "delete_selected",
     ]
 
+    fieldsets = (
```
marcoacierno, line 318: weird order where "Manage" appears before "The grant"?

backend/grants/admin.py
```diff
@@ -317,12 +310,59 @@ class GrantAdmin(ExportMixin, AdminUsersMixin, SearchUsersMixin):
...
+                    "accommodation_amount",
+                    "total_amount",
+                    "applicant_message",
+                    "applicant_reply_sent_at",
+                    "applicant_reply_deadline",
+                )
+            },
+        ),
+        (
+            "About the Applicant",
```
marcoacierno, line 337: same, the order seems off

when I read a grant I imagine I should see

The grant I am looking at -> Info about the person -> Actions to modify it

backend/grants/admin.py
```diff
@@ -298,7 +288,10 @@ class GrantAdmin(ExportMixin, AdminUsersMixin, SearchUsersMixin):
         "applicant_reply_sent_at",
         "applicant_reply_deadline",
     )
-    readonly_fields = ("email",)
+    readonly_fields = (
+        "email",
```
marcoacierno, line 292: why are we still showing this field? I would  just remove it

backend/grants/admin.py
```diff
@@ -278,6 +266,8 @@ class Meta:
             "travelling_from",
             "country_type",
             "applicant_message",
+            "applicant_reply_sent_at",
+            "applicant_reply_deadline",
```
marcoacierno, line 270: interesting that each grant has it is own deadline instead of being a global thing

estyxx: Because you can send the email separately, for example: if we receive more founds, we will send other emails with a different deadline of course, in 14 days from the date the email is sent. 
This is the deadline for the applicant to reply and confirm the grant, that is different to the `Waiting List Deadline` I added in the deadlines, that one is global and is for telling them when they will receive an update.

backend/grants/admin.py
estyxx: it's empty now, so we can remove it

backend/grants/admin.py
estyxx: Because in the admin we are editing this information mostly, so it's better to have this section first in my opinion. And Django automatically will show the `full_name` of the person in bold at the top, so you know what grant you are editing already.

estyxx: if you think about we have kind the same order now in production, without the fieldsets: 
Screenshot 2023-02-11 at 12 46 07

The new admin, you see the full name here: 
Screenshot 2023-02-11 at 12 45 35

marcoacierno: The default order in production is a "random" one so I wouldn't necessarily use it as an example of "good UX to avoid mistakes"

If you think your order is ok and works then feel free to leave it as it is

marcoacierno (approved):

PR #3374, opened 2023-03-16
Grants: Add Waiting LIst update email

estyxx:


backend/grants/admin.py
```diff
@@ -374,6 +401,41 @@ def user_display_name(self, obj):
...
+        qs = super().get_queryset(request)
+        if not self.speaker_ids:
+            # returns a list...
+            conference_id = qs.values_list("conference_id").first()
+            self.speaker_ids = [
+                i["submission__speaker_id"]
+                for i in ScheduleItem.objects.filter(
+                    conference__id__in=conference_id
+                ).values("submission__speaker_id")
+                if i["submission__speaker_id"]
```
marcoacierno, line 434: why not filter on the query itself? `submission__speaker_id__isnull=False`?

backend/grants/admin.py
```diff
@@ -374,6 +401,41 @@ def user_display_name(self, obj):
...
+    def is_speaker(self, obj):
+        if obj.user_id in self.speaker_ids:
+            return "🗣️"
+        return ""
+
+    def get_queryset(self, request):
+
+        qs = super().get_queryset(request)
+        if not self.speaker_ids:
+            # returns a list...
```
marcoacierno, line 427:
```suggestion
```

backend/grants/admin.py
```diff
@@ -277,10 +300,13 @@ class Meta:
 
 @admin.register(Grant)
 class GrantAdmin(ExportMixin, AdminUsersMixin, SearchUsersMixin):
+    speaker_ids = []
```
marcoacierno, line 303: is this safe knowing that the admin object is a singleton?

backend/grants/admin.py
```diff
@@ -374,6 +401,41 @@ def user_display_name(self, obj):
...
+        description="S",
+    )
+    def is_speaker(self, obj):
+        if obj.user_id in self.speaker_ids:
+            return "🗣️"
+        return ""
+
+    def get_queryset(self, request):
+
+        qs = super().get_queryset(request)
```
marcoacierno, line 425: nitpick: I imagine another way to do it would be to use `.annotate(is_speaker=Subquery`) but I am not sure it would be more efficient or not than merging in python

backend/grants/admin.py
```diff
@@ -238,6 +240,27 @@ def send_grant_reminder_to_waiting_for_confirmation(modeladmin, request, queryse
...
+        ),
+    )
+
+    for grant in queryset:
+        if grant.applicant_reply_sent_at is not None:
+            messages.warning(
+                request,
+                f"Reply email for {grant.name} was already sent! Skipping.",
+            )
+            return
```
marcoacierno, line 258: is the return correct here? I think you want `continue` otherwise it stops everything

backend/grants/admin.py
```diff
@@ -238,6 +240,27 @@ def send_grant_reminder_to_waiting_for_confirmation(modeladmin, request, queryse
...
+def send_reply_email_waiting_list_update(modeladmin, request, queryset):
+    queryset = queryset.filter(
+        status__in=(
+            Grant.Status.waiting_list,
+            Grant.Status.waiting_list_maybe,
+        ),
+    )
+
+    for grant in queryset:
+        if grant.applicant_reply_sent_at is not None:
```
marcoacierno, line 253: uhm is this check correct? I imagine we already sent an email so this field will not be None for everyone

PR #3383, opened 2023-03-17
Grants Admin: Add create vouchers on pretix action

estyxx:


backend/grants/admin.py
```diff
@@ -240,6 +243,79 @@ def send_reply_email_waiting_list_update(modeladmin, request, queryset):
...
+        send_grant_voucher_email(grant)
+        count = count + 1
+
+    messages.success(request, f"{count} Voucher emails scheduled!")
+
+
+def _generate_voucher_code() -> str:
+    charset = list("ABCDEFGHKLMNPQRSTUVWXYZ23456789")
+    random_string = get_random_string(length=20, allowed_chars=charset)
+    return f"GRANT-{random_string}"
```
marcoacierno, line 267: I am sure we have a similar function for speakers, we should refactor and share it in the two places?

```suggestion
def _generate_voucher_code(prefix: str) -> str:
    charset = list("ABCDEFGHKLMNPQRSTUVWXYZ23456789")
    random_string = get_random_string(length=20, allowed_chars=charset)
    return f"{prefix}-{random_string}"
```

marcoacierno (approved):
left 1 minor comment

PR #3384, opened 2023-03-19
Missing text in the italian language

alex9446:
## Why
On the workshop pages, the space left info is rendered as follows:

immagine


marcoacierno:
Amazing, thanks a lot!

PR #3500, opened 2023-05-07
Badge scanning API

patrick91:
This PR adds the API for scanning badges :)

# TODO:

- [x] Add models
- [x] Store scans on db
- [x] Deduplicate scans (onle one row per scanner and "scannee")?

# For other PR:

- [ ] Export scans on a CSV
- [ ] Add permission for scanning badges (staff[1] and sponsors only)
- [ ] Add API to update scan with note
- [ ] List scans for users on their profile

[1] Staff for testing


backend/api/badge_scanner/schema.py
```diff
@@ -0,0 +1,95 @@
...
+class ScanError:
+    message: str
+
+
+@strawberry.input
+class ScanBadgeInput:
+    url: str
+    conference_code: str
+
+    url_regex = re.compile(r"^https://pycon\.it/b/([\w\d-]+)$")
```
marcoacierno, line 26: we should move this outside so the RegEx will be compiled only once, instead of being compiled for every request (i.e every time ScanBadgeInput is created)

backend/api/badge_scanner/schema.py
```diff
@@ -0,0 +1,95 @@
...
+@strawberry.type
+class Attendee:
+    full_name: str
+    email: str
+
+
+@strawberry.type
+class BadgeScan:
+    attendee: Attendee
+    notes: Optional[str]
```
marcoacierno, line 54:
```suggestion
    notes: str
```

backend/api/badge_scanner/schema.py
```diff
@@ -0,0 +1,95 @@
...
+        match = self.url_regex.match(self.url)
+
+        assert match
+
+        return match.group(1)
+
+
+@strawberry.type
+class Attendee:
+    full_name: str
```
marcoacierno, line 47:
```suggestion
    fullname: str
```

I need to double-check, but I think in all places we are now preferring `fullname` instead of `fullName`

patrick91: CleanShot 2023-05-10 at 10 46 56@2x

we use fullName more, happy to change it to fullname though

backend/api/badge_scanner/schema.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@strawberry.type
+class BadgeScannerMutation:
+    @strawberry.mutation(
+        permission_classes=[IsAuthenticated],
+    )
+    def scan_badge(
+        self, info: Info[Any, None], input: ScanBadgeInput
+    ) -> Union[BadgeScan, ScanError]:
```
marcoacierno, line 64: since you added the future import and are using it above, you should use the pipe operator here as well to be consistent

```suggestion
    ) -> BadgeScan | ScanError:
```

backend/api/badge_scanner/schema.py
marcoacierno: it would look better outside regardless I think 🤔

backend/api/badge_scanner/schema.py
marcoacierno: Is this how many times the field is called or the number of time a field named `fullName` exists in the API? I think in all my APIs I have used `fullname` now

backend/api/badge_scanner/schema.py
```diff
@@ -0,0 +1,94 @@
...
+
+        return None if match else ScanError(message="URL is not valid")
+
+    @property
+    def order_position_id(self) -> str:
+        match = self.url_regex.match(self.url)
+
+        assert match
+
+        return match.group(1)
```
marcoacierno, line 41: This should be changed to decode the ID as well as we won't have the raw ID in the URL

e.g it won't be `pycon.it/b/132` but `pycon.it/b/abc`

We should do something similar to what we did for submissions, use hashids and the `decode_hashid` and `encode_hashid` functions (it should be fine to still use the secret key as salt, we can think of improving it later)

backend/api/badge_scanner/schema.py
patrick91: I prefer it being scoped to that class 😊

unless we move some of this logic somewhere else

backend/api/badge_scanner/schema.py
```diff
@@ -0,0 +1,94 @@
...
+
+        if not conference:
+            return ScanError(message="Conference not found")
+
+        if error := input.validate():
+            return error
+
+        data = pretix.get_order_position(conference, input.order_position_id)
+
+        user = get_user_by_email(data["attendee_email"])
```
marcoacierno, line 74: `user` is a bit confusing: who is that? the person scanning the badge or the person scanned?

I would name it like this so it is clear: 

```suggestion
        scanned_user = get_user_by_email(data["attendee_email"])
```

backend/api/badge_scanner/schema.py
patrick91: how many times the field is called

patrick91:
Merging, the failing test is unrelated :)

PR #3543, opened 2023-05-21
Add lifecycle to delete badge scans after one day

patrick91:

### DIFF
--- a/infrastructure/applications/pycon_backend/s3.tf
+++ b/infrastructure/applications/pycon_backend/s3.tf
@@ -10,3 +10,18 @@ resource "aws_s3_bucket" "backend_media" {
     max_age_seconds = 3000
   }
 }
+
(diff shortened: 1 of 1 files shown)


patrick91:
this is only available in newer versions of terraforms unfortunately

marcoacierno:
I don't think we should just delete the files from S3, as we will end up with invalid rows in our database with files that point to nothing. I think we can leave it as it is for now and improve it later?

patrick91:
> I don't think we should just delete the files from S3, as we will end up with invalid rows in our database with files that point to nothing. I think we can leave it as it is for now and improve it later?

you won't be able to access older scans unless you go to the admin, so I think it is fine :)

another option is to have a cron to delete scans from admin and s3, but this solution seemed fine enough since it is a built in of s3, sad we can't use it via terraform for now :(

PR #3629, opened 2023-11-30
Add invoice fields

patrick91:
Before merging this I need to:

- [x] Add pretix plugin that has the new API to store SDI and PEC
- [x] Call the API
- [x] Add tests for that


patrick91:
I haven't checked but I think tslint is failing because it doesn't check agains the repo version it checks agains prod, I might make two PRs for this 😊

PR #3654, opened 2023-12-29
Refactor Grant Summary implementation in Admin

estyxx:
## Why

This PR introduces a significant refactor of the `Grant`  summary view in the Django admin panel. The previous implementation, which involved an auxiliary model GrantRecap and a convoluted override of the changelist_view, has been replaced with a cleaner, more maintainable, and extendable approach.


backend/grants/admin.py
```diff
@@ -548,88 +552,137 @@ def save_form(self, request, form, change):
...
-            "confirmed": counter.get(Grant.Status.confirmed, 0),
+        ]
+        return custom_urls + urls
+
+    def summary_view(self, request):
+        """
+        Custom view for summarizing Grant data in the Django admin.
+        Aggregates data by country and status, and applies request filters.
+        """
+        # Initialize statuses
```
marcoacierno, line 571: I don't think we need those comments? "Initialize statuses" "Apply filters to the Grant queryset" etc

backend/grants/admin.py
```diff
@@ -548,88 +552,137 @@ def save_form(self, request, form, change):
...
+        statuses = Grant.Status.choices
+
+        # Apply filters to the Grant queryset
+        filter_params = request.GET.dict()
```
marcoacierno, line 575: uhm this is probably fine because it is in admin, but I am not a big fan of just forwarding whatever is in the GET parameters to filter 🤔

backend/grants/templates/admin/grants/grant_summary.html
```diff
@@ -0,0 +1,163 @@
...
+                <th>Continent</th>
+                <th>Country</th>
+                {% for status in statuses %}<th>{{ status.1|title }}</th>{% endfor %}
+            </tr>
+        </thead>
+        <tbody>
+            <tbody>
+                {% for key, counts in summary.items %}
+                    {% with continent=key.0 country=key.1 %}
+                        {% ifchanged continent %}
```
marcoacierno, line 98: 💯

backend/grants/templates/admin/grants/grant_summary.html
```diff
@@ -0,0 +1,163 @@
...
+                        <td>{{ genders|get_item:gender|default:"Not Specified" }}</td>
+                        <td></td>
+                        {% for status in statuses %}<td>{{ gender_data|get_item:status.0|default:"0" }}</td>{% endfor %}
+                    </tr>
+                {% endfor %}
+            </tbody>
+        </table>
+        <script>
+        document.addEventListener('DOMContentLoaded', function() {
+            var checkboxes = document.querySelectorAll('.toggle-column');
```
marcoacierno, line 146: `const`?

backend/grants/templates/admin/grants/grant_summary.html
```diff
@@ -0,0 +1,163 @@
...
+            checkboxes.forEach(function(checkbox) {
+                checkbox.addEventListener('change', function() {
+                    var countryColumn = document.getElementById('grantsByCountry').querySelectorAll('td:nth-child(' + checkbox.dataset.column + '),th:nth-child(' + checkbox.dataset.column + ')');
```
marcoacierno, line 149: why are you using both `getElementById` and `querySelectorAll`? can you not just use `querySelectorAll`?

marcoacierno: or maybe using JS template strings would look better

backend/grants/templates/admin/grants/grant_summary.html
```diff
@@ -0,0 +1,163 @@
...
+        }
+    </style>
+{% endblock %}
+{% block content %}
+    <div class="back-button">
+        <a href="{% url 'admin:grants_grant_changelist' %}?{{ request.GET.urlencode }}"
+           class="button">Back to List</a>
+    </div>
+    <h1>Grant Summary</h1>
+    <!-- Display Active Filters -->
```
marcoacierno, line 60:
```suggestion
```

backend/grants/templates/admin/grants/grant_summary.html
```diff
@@ -0,0 +1,163 @@
...
+        document.addEventListener('DOMContentLoaded', function() {
+            var checkboxes = document.querySelectorAll('.toggle-column');
+            checkboxes.forEach(function(checkbox) {
+                checkbox.addEventListener('change', function() {
+                    var countryColumn = document.getElementById('grantsByCountry').querySelectorAll('td:nth-child(' + checkbox.dataset.column + '),th:nth-child(' + checkbox.dataset.column + ')');
+                    countryColumn.forEach(function(cell) {
+                        cell.style.display = checkbox.checked ? '' : 'none';
+                    });
+
+                    var genderColumn = document.getElementById('grantsByGender').querySelectorAll('td:nth-child(' + checkbox.dataset.column + '),th:nth-child(' + checkbox.dataset.column + ')');
```
marcoacierno, line 154: using nth-child feels very hacky. Could we maybe use data attributes?

What are you trying to do here?

backend/grants/admin.py
```diff
@@ -361,6 +363,8 @@ class Meta:
 
 @admin.register(Grant)
 class GrantAdmin(ExportMixin, admin.ModelAdmin):
+    change_list_template = "admin/grants/grant/change_list.html"
```
marcoacierno, line 366: is this django default path? I don't think you need this line if so?

backend/grants/admin.py
```diff
@@ -548,88 +552,137 @@ def save_form(self, request, form, change):
...
+            gender = data["user__gender"] if data["user__gender"] else ""
+            status = data["status"]
+            total = data["total"]
+            gender_summary[gender][status] += total
+
+        return gender_summary
+
+    def _format_filters_for_display(self, filter_params):
+        """
+        Formats raw filter keys to user-friendly names.
```
marcoacierno, line 661: this comment doesn't look useful  as it just explains what's the code after it? 🤔

backend/grants/admin.py
estyxx: I moved this logic in `_filter_and_format_grants` to filter only by the allowed filters

backend/grants/admin.py
estyxx: I changed this method and updated the description, let me know if now is more useful :)

backend/grants/templates/admin/grants/grant_summary.html
estyxx: Add the option to show and hide status columns

estyxx: I changed to use the data-column

backend/grants/templates/admin/grants/grant_summary.html
```diff
@@ -0,0 +1,194 @@
...
+                {% for status in statuses %}
+                    <td class="status-column-{{ forloop.counter }}">{{ gender_data|get_item:status.0|default:"0" }}</td>
+                {% endfor %}
+            </tr>
+        {% endfor %}
+    </tbody>
+</table>
+<script>
+  document.addEventListener('DOMContentLoaded', function() {
+      // This script allows users to toggle the visibility of columns in the table
```
marcoacierno, line 176: is this comment useful?

marcoacierno (approved):

marcoacierno:
There are still various comments that explain what the code below does, which I don't believe are helpful, but I will leave it to you to decide what you think makes sense to keep and what not.

PR #3664, opened 2024-01-02
Monitoring

marcoacierno:


marcoacierno:
I need to spend more time thinking about this. Ideally, what we care about most is visibility into what is happening on Celery (e.g. are tasks being processed?)

PR #3686, opened 2024-01-18
CMS Headless Preview

marcoacierno:
## ToDo:

- [x] Replace the hack `conference code <-> site hostname` hack. Site hostname should be the actual hostname of the frontend (so 2024.pycon.it) while the conference code mapping should  be done via a custom site model. We should rename VercelFrontendSettings to be PyCon specific
- [x] We use a fork of wagtail headless because we want the preview url to contain the page path and not just go to the site root - look if there is an alternative to a fork or cleanup the forked repo/package if not (https://github.com/pythonitalia/wagtail-headless-preview)
- [x] As above ^ handle when you create new pages. Not a big issue as you could solve it by saving it as a draft to start seeing the preview.
- [x] Revist the FE approach: Right now it is inside BlocksRenderer. This was a quick and dirty way to do it. Can we do it better? Can we make this work for News articles as well? 
    - [x] NextJS had some Preview function, we should look into it: https://nextjs.org/docs/pages/building-your-application/configuring/draft-mode Let's look into it but if it is too overkill I am fine with using the current solution
- [x] Tests for page preview
- [x] Test to make sure we don't return draft revisions

## Unrelated (mostly):

- [ ] We should start looking at Menus. Right now they are still in our Backend and it is not ideal. 
- [ ] Can we get rid of the Homepage Page type?


frontend/next.config.js
```diff
@@ -28,6 +28,10 @@ module.exports = withSentryConfig({
             key: "X-Frame-Options",
             value: "DENY",
           },
+          {
+            key: "Content-Security-Policy",
+            value: "frame-ancestors localhost:8000;",
```
marcoacierno, line 33: update this

backend/api/cms/news/queries/news_articles.py
```diff
@@ -6,16 +6,18 @@
 
 @strawberry.field
 def news_articles(hostname: str, language: str) -> list[NewsArticle]:
-    site = Site.objects.filter(hostname=hostname).first()
+    hostname, port = hostname.split(":") if ":" in hostname else (hostname, 80)
```
marcoacierno, line 9: write tests for this

backend/api/cms/tests/news/test_queries.py
```diff
@@ -110,6 +121,87 @@ def test_get_news_article(
...
+    parent = GenericPageFactory()
+    article_1 = NewsArticleFactory(
+        title="Article 1",
+        parent=parent,
+        owner=user,
+        slug="slug",
+        first_published_at=None,
+        live=False,
+    )
+    SiteFactory(hostname="pycon", port=80, root_page=parent)
```
estyxx, line 184: Why don't we use pytest-factoryboy? `site_factory` as a fixtre...
Are we using https://github.com/wagtail/wagtail-factories ? 
Probably some factories are not registered, unfortunately, but we can just register on our end...

backend/api/cms/news/queries/news_article.py
```diff
@@ -6,7 +6,8 @@
 
 @strawberry.field
 def news_article(hostname: str, slug: str, language: str) -> NewsArticle | None:
-    site = Site.objects.filter(hostname=hostname).first()
+    hostname, port = hostname.split(":") if ":" in hostname else (hostname, 80)
```
estyxx, line 9: Do we have the request object somewhere? 
Maybe we can use `Site.find_for_request(request)` https://docs.wagtail.org/en/stable/reference/pages/model_reference.html#wagtail.models.Site.find_for_request ?

estyxx:
> We use a fork of wagtail headless because we want the preview url to contain the page path and not just go to the site root - look if there is an alternative to a fork or cleanup the forked repo/package if not (https://github.com/pythonitalia/wagtail-headless-preview)

How did you solve it? I didn't see any forks... 🤔

backend/api/cms/tests/news/test_queries.py
marcoacierno: I am migrating away from using factories as fixtures as the test code ends up unnecessarily messy when done. Look at any test code that imports 5 factories, and see it has a long list of parameters.

It gets even messier when you want to maybe add params to the test, and now you need to see where your params would get added.

In general I prefer importing the factory once and using it everywhere it is needed

backend/api/cms/news/queries/news_article.py
marcoacierno: Unfortunately, I don't believe it would work with our current setup as we have a hacky middleware that forces the Host of the request to be pycon.it (hopefully, one day, we will get rid of it), so it wouldn't be able to match with 2023.pycon.it / 2024.pycon.it

Once we remove the middleware, it could be an option, but it would limit our API as you won't be able to request any of our CMS data anymore unless you pass through a layer that makes sure the API hostname matches the CMS hostname. I am not against it, as in terms of security, this endpoint allows everyone to fetch any page for any site we have in our CMS, which is not ideal

marcoacierno:
> > We use a fork of wagtail headless because we want the preview url to contain the page path and not just go to the site root - look if there is an alternative to a fork or cleanup the forked repo/package if not ([pythonitalia/wagtail-headless-preview](https://github.com/pythonitalia/wagtail-headless-preview))
> 
> How did you solve it? I didn't see any forks... 🤔

@estyxx  At the end I just realised I didn't need the fork (I deleted it, but it is still on PyPi) and just was able to integrate Vercel by changing the options here: https://github.com/pythonitalia/pycon/pull/3686/files#diff-46f4b62543c18a2713ac153216287b3688ac1fffa0a464111e4b52a276da0489R361

PR #3687, opened 2024-01-21
Grants admin: Add column for Confirmed and Proposed speakers

estyxx:
## Why

To know if the Grant applicant has proposed a talk or not (at this stage the talks are not in the schedule yet). 
I modified the existing column to be "is_confirmed_talk" and added a new one "is_proposed_talk". 
As this column is just an emoji I added a tooltip custom in the admin so the user can go hover the title column and see the extended name of the column...    Django unfortunately doesn't have this functionality by default.

### DIFF
--- a/backend/grants/templates/admin/grants/grant/change_list.html
+++ b/backend/grants/templates/admin/grants/grant/change_list.html
@@ -6,24 +6,7 @@
     </li>
     {{ block.super }}
 {% endblock %}
-{% block extrastyle %}
-    <style>
-  th[title]:hover:after {
-      content: attr(title);
-      padding: 4px 8px;
-      color: #fff;
-      position: absolute;
-      left: 50%;
-      bottom: 100%;
-      white-space: nowrap;
-      z-index: 1000;
-      background-color: #333;
-      border-radius: 6px;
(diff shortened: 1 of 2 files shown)


backend/grants/templates/admin/grants/grant/change_list.html
```diff
@@ -1,7 +1,39 @@
...
     <li>
-        <a href="{% url 'admin:grants-summary' %}?{{ request.GET.urlencode }}" class="button">Grant Summary</a>
+        <a href="{% url 'admin:grants-summary' %}?{{ request.GET.urlencode }}"
+           class="button">Grant Summary</a>
     </li>
     {{ block.super }}
 {% endblock %}
+{% block extrastyle %}
+    <style>
+  th[title]:hover:after {
```
estyxx, line 11: Questi stili li ho copiati pari pari da chatGPT senza pensarci troppo...

estyxx: CleanShot 2024-01-21 at 21 05 50@2x
E viene fuori cosi'

marcoacierno (approved):

PR #3705, opened 2024-01-30
Grants: update plain API to create new chat

estyxx:


backend/integrations/plain.py
```diff
@@ -155,19 +137,27 @@ def _send_chat(customer_id: str, title: str, message: str):
...
-    change_customer_status(customer_id)
-    _send_chat(customer_id, title, message)
+    thread_id = _create_thread(customer_id, title, message)
+
+    try:
+        grant = Grant.objects.get(user=user)
+        grant.plain_thread_id = thread_id
+        grant.save()
+    except Grant.DoesNotExist:
+        logger.error("Coudn't find the grant for: %s", user.user_id)
```
estyxx, line 163: Don't think someone can send a message without the grant really but, just in case...

backend/grants/models.py
```diff
@@ -189,6 +189,12 @@ class ApprovedType(models.TextChoices):
         _("applicant reply deadline"), null=True, blank=True
     )
     applicant_message = models.TextField(_("applicant message"), null=True, blank=True)
+    plain_thread_id = models.CharField(
+        _("Plain threadID"),
+        max_length=50,
```
marcoacierno, line 194: Why the max length? Is this 50 correct?

backend/integrations/plain.py
```diff
@@ -155,19 +137,27 @@ def _send_chat(customer_id: str, title: str, message: str):
...
+    return thread_id
 
 
 def send_message(user: User, title: str, message: str):
     customer_id = create_customer(user)
-    change_customer_status(customer_id)
-    _send_chat(customer_id, title, message)
+    thread_id = _create_thread(customer_id, title, message)
+
+    try:
```
marcoacierno, line 158: This is a bit of mix of concerns, the plain integration doesn't/shouldnt know anything about the grant

Say we want to re-use this function to send some other messages to Plain

backend/grants/models.py
estyxx: I put a bit longer. think it's 29 really

marcoacierno (approved):

backend/integrations/plain.py
estyxx: yeah I see

PR #3712, opened 2024-02-01
Grants: email copy updates

estyxx:


backend/grants/admin.py
```diff
@@ -160,28 +160,14 @@ def _check_amounts_are_not_empty(grant: Grant, request):
...
 
-    if (
-        grant.grant_type
-        in (
-            Grant.ApprovedType.ticket_accommodation,
-            Grant.ApprovedType.ticket_travel_accommodation,
-        )
-        and grant.accommodation_amount is None
-    ):
+    if grant.has_approved_accommodation() and grant.accommodation_amount is None:
```
marcoacierno, line 163: not for this PR, but we have a very similar logic in the Grant save method when it calculates the amounts to set. We could refactor it to use those 2 new methods

email-templates/src/templates/grants-approved.html
```diff
@@ -0,0 +1,191 @@
...
+                                                receipts, like plane/bus/train tickets, sent to us before the
+                                                conference
+                                            </li>
+                                        @{{/if}}
+                                        @{{#if hasApprovedAccommodation}}
+                                            <li>
+                                                <span class="font-semibold">🛏️ Accommodation:</span>
+                                                <br />
+                                                A bed in a shared room at the "Grand Hotel Mediterraneo," where
+                                                the conference take place. This accommodation is available from
```
marcoacierno, line 51:
```suggestion
                                                the conference takes place. This accommodation is available from
```

marcoacierno (approved):

PR #3720, opened 2024-02-13
Alert about different durations, emails updates on Schedule

marcoacierno:


email-templates/src/templates/submission-schedule-time-change.html
```diff
@@ -1,5 +1,5 @@
 ---
-title: "[PyCon Italia 2022] Your submission has a new time!"
+title: "[PyCon Italia 2024] Your submission has a new time!"
```
estyxx, line 2: What does this title does? Though was for rendering in the localhost rendering but don't think is working...

marcoacierno: I think it would be used if we used this node process to send the emails

PR #3754, opened 2024-03-05
Grants admin: Order by user's full name

estyxx:

### DIFF
--- a/backend/grants/admin.py
+++ b/backend/grants/admin.py
@@ -533,9 +533,7 @@ class GrantAdmin(ExportMixin, ConferencePermissionMixin, admin.ModelAdmin):
         ),
     )
 
-    @admin.display(
-        description="User",
-    )
+    @admin.display(description="User", ordering="user__full_name")
     def user_display_name(self, obj):
         if obj.user_id:
             return obj.user.display_name


backend/grants/admin.py
```diff
@@ -533,9 +533,7 @@ class GrantAdmin(ExportMixin, ConferencePermissionMixin, admin.ModelAdmin):
         ),
     )
 
-    @admin.display(
-        description="User",
-    )
+    @admin.display(description="User", ordering="user__full_name")
```
estyxx, line 536: In the admin I cannot order by `user.display_name` unfortunately nor I can do `user__full_name,user__name` as order option, so I choose to order by full_name... 
The default sorting is the primary key btw

marcoacierno (approved):

backend/grants/admin.py
marcoacierno: had no idea this existed, does it mean that when you click on "User" it will sort on full_name? that's cool

marcoacierno: you can also set the default ordering to be honest, e..g: order by conference and name

so they are grouped together

PR #3775, opened 2024-03-18
Add gender filter to Grants Admin

estyxx:

### DIFF
--- a/backend/grants/admin.py
+++ b/backend/grants/admin.py
@@ -420,6 +420,7 @@ class GrantAdmin(ExportMixin, ConferencePermissionMixin, admin.ModelAdmin):
         "country",
         "is_proposed_speaker",
         "is_confirmed_speaker",
+        "emoji_gender",
         "conference",
         "status",
         "approved_type",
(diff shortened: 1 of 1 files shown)


backend/grants/admin.py
```diff
@@ -443,14 +444,15 @@ class GrantAdmin(ExportMixin, ConferencePermissionMixin, admin.ModelAdmin):
         "status",
         "country_type",
         "occupation",
-        "grant_type",
+        "approved_type",
```
estyxx, line 447: "grant_type" (Speaker | Diversity | Unemployed) don't think we actually use it as now there are the more precise filters "IsProposedSpeakerFilter". 
There was missing instead `approved_type` that is useful instead.

marcoacierno: This is used during reviews I think?

marcoacierno (approved):

backend/grants/admin.py
estyxx: mmm not really if I remember correctly... And now there is the review system so...

PR #3830, opened 2024-05-02
New email speakers system

marcoacierno:


marcoacierno:
We are working on designing a more generic approach to emails that will cover speakers and grants

PR #3879, opened 2024-07-21
Remove applicant_message from grants

estyxx:
## Why

The `applicant_message` contains sensitive information that we also have in Plain. Initially was introduced to save the message if the grantee needs more information in order to accept the grant and just in case something happen with the Plain API, so was meant to be a temporary thing to be removed later. Also now we have the VISA form in place we don't need people to write here stuff about the visa. They can just send us a mail.


marcoacierno:
If we merge this and remove the fields from the model we will lose the data stored there. Is there anything important we don't want to lose in there?

estyxx:
> If we merge this and remove the fields from the model we will lose the data stored there. Is there anything important we don't want to lose in there?

I don't think so. Everything should also be on Plain if we need it. The applicant_message was just a backup, and plain_thread_id was to continue the conversation if they edited the message... 
It's also full of very sensitive information like passports etc. the conference is over we don't need these anymore

marcoacierno (changes requested):
Vercel build is failing because of `Type error: Property 'message' does not exist on type 'StateValues<GrantReplyFrom>'.` can you fix it?

estyxx:
> Vercel build is failing because of `Type error: Property 'message' does not exist on type 'StateValues<GrantReplyFrom>'.` can you fix it?

I should have fixed: I removed the message field from the input form. But I have no access to Vercel anymore, don't know why is failing now...

marcoacierno:
> Vercel build is failing because of Type error: Property 'message' does not exist on type 'StateValues<GrantReplyFrom>'. can you fix it?
> I should have fixed: I removed the message field from the input form. But I have no access to Vercel anymore, don't know why is failing now...

If you run the build locally `pnpm build` it should show you. At some point we should also add a CI step that runs typescript

marcoacierno:
Right now it should be failing because of the mismatch between prod and local in the types. Something I am trying to do more often is to do 2 separate PRs one for BE and one for FE to make sure we don't have surprise issues when merging.

Can you try running the build locally `pnpm build` and let me know if it builds ok? @estyxx

PR #3966, opened 2024-08-08
Add BE support for given/family attendee name

marcoacierno:
## ToDo

- [x] Fix/Implement name validation
- [x] Double-check error reporting
- [x] Migrate update_attendee_ticket to the new errors api
- [x] Make update and create order field consistent: name both `attendeeName`

### DIFF
--- a/backend/pretix/__init__.py
+++ b/backend/pretix/__init__.py
@@ -256,12 +256,18 @@ class InvoiceInformationErrors:
 
 
 @strawberry.type
+class CreateOrderTicketErrors:
+    attendee_name: list[str] = strawberry.field(default_factory=list)
+
+
+@strawberry.type
 class CreateOrderErrors(BaseErrorType):
     @strawberry.type
     class _CreateOrderErrors:
         invoice_information: InvoiceInformationErrors = strawberry.field(
             default_factory=InvoiceInformationErrors
         )
+        tickets: list[CreateOrderTicketErrors] = strawberry.field(default_factory=list)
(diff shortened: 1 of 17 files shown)


backend/pretix/__init__.py
```diff
@@ -276,12 +282,18 @@ class CreateOrderTicketAnswer:
...
     ticket_id: str
-    attendee_name: str
+    attendee_name: AttendeeNameInput
     attendee_email: str
     variation: Optional[str] = None
     answers: Optional[List[CreateOrderTicketAnswer]] = None
     voucher: Optional[str] = None
 
+    def validate(self, errors: CreateOrderErrors) -> CreateOrderErrors:
+        if not self.attendee_name.validate():
```
marcoacierno, line 292: we should add the same check for emails

right now I believe we rely on pretix api to reject the email

backend/pretix/__init__.py
```diff
@@ -399,11 +411,18 @@ class CreateOrderInput:
...
 
     def validate(self) -> CreateOrderErrors:
         errors = CreateOrderErrors()
-        self.invoice_information.validate(errors)
+
+        with errors.with_prefix("invoice_information"):
+            self.invoice_information.validate(errors)
+
+        for index, ticket in enumerate(self.tickets):
+            with errors.with_prefix(f"tickets.{index}"):
```
marcoacierno, line 423: maybe this is nicer?

```python
errors.with_prefix("tickets", index)
errors.with_prefix("tickets", index, "another")
```

instead of manually specifying `.` and strings?

Issue #3929, opened 2024-08-03
Move Grant Summary view to Conference admin for restricted access


estyxx:
Currently, the Grant Admin includes a "summary" view to recap the budget spent and how grants are assigned by country, gender, etc., accessible via [Grant Summary](https://admin.pycon.it/admin/grants/grant/summary/).

This view is accessible to anyone with access to the grant model. We must restrict access to ensure only authorized admin users for a specific conference can view this sensitive data.

We need to move the admin path from the Grant Admin to the Conference Admin. This change will restrict access based on conference permissions.

Remove the custom URL from the Grant admin `grants/admin.py` 
    ```python
    path(
        "summary/",
        self.admin_site.admin_view(self.summary_view),
        name="grants-summary",
    ),
    ```
    
 Add the custom URL with the object_id in the Conference admin `conference/admin.py`:
    ```python
    path(
        "<int:object_id>/grants-summary/",
        self.admin_site.admin_view(self.grants_summary),
        name="grants_summary",
    ),
    ```

PR #4005, opened 2024-08-18
Move grant summary under Conference

estyxx:
## How to test it

#3929


backend/grants/templates/admin/grants/grant_summary.html
```diff
@@ -14,11 +19,11 @@
             padding: 6px 12px;
         }
 
-        .back-button a.button:hover {
-            background-color: #e6e6e6;
+        .approved, .waiting_for_confirmation, .confirmed {
```
estyxx, line 22: I added to highlight those statuses that contribute to the total

estyxx:
I don't know how to fix the tests: the `messages` are correctly mock `mock_messages = mocker.patch("conferences.admin.conference.messages")` so don't understand why it's complaining...

marcoacierno:
> I don't know how to fix the tests: the `messages` are correctly mock `mock_messages = mocker.patch("conferences.admin.conference.messages")` so don't understand why it's complaining...

@estyxx Looking at the code you moved the actions (what the tests are testing, e.g. `send_voucher_via_email`) in `conferences/admin/actions.py` meaning that you should mock `conferences.admin.actions.messages` and not `conferences.admin.conference.messages`.

backend/grants/models.py
```diff
@@ -37,6 +37,12 @@ class Status(models.TextChoices):
         Status.waiting_list_maybe.value,
     ]
 
+    BUDGET_STATUSES = [
```
marcoacierno, line 40: what does `budget` mean here?

backend/grants/summary.py
```diff
@@ -0,0 +1,119 @@
+from django.db.models import Count, Sum
+from conferences.models.conference import Conference
+from helpers.constants import GENDERS
+from countries import countries
+from .models import Grant
+
+
+class GrantSummary:
```
marcoacierno, line 8: it might be good to cover this class with tests if we don't have them already? how do we make sure the calculations are correct?

marcoacierno (changes requested):
The grants-recap screen crashes because it can't find grants-summary as the name was changed to `grants_summary`

backend/grants/models.py
estyxx: I moved it in the grant summary, as it's just for the budget calculations. 
We want to count only the "approved" / confirmed or those grant in `waiting_for_confirmtion`  in the allocated budget.

estyxx:
@marcoacierno I added/fixed the tests fyi

estyxx:
@marcoacierno  I applied your suggestions, do you think we can merge this?

marcoacierno (approved):

PR #4026, opened 2024-09-01
Move off lambda and to ECS

marcoacierno:


marcoacierno:
I am unhappy with the terraform code. It looks rather messy, and the single server setup had some instability and random issues, so it makes more sense for me to start from scratch maybe looking into CDK or using CDK to Terraform and similar to have a more clean code.

PR #4037, opened 2024-09-06
Rename shop to store

patrick91:

### DIFF
--- a/infrastructure/global/domains/python_it/records.tf
+++ b/infrastructure/global/domains/python_it/records.tf
@@ -400,129 +400,129 @@ resource "aws_route53_record" "social_python_it_fly_challange" {
 
 # Fourth Wall Shop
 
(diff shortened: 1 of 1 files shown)


infrastructure/global/domains/python_it/records.tf
```diff
@@ -400,129 +400,129 @@ resource "aws_route53_record" "social_python_it_fly_challange" {
...
   type    = "CNAME"
-  records = ["shop.python.it."]
+  records = ["store.python.it."]
   ttl     = "3600"
 }
 
-resource "aws_route53_record" "em_fw_support_shop_python_it_cname" {
+resource "aws_route53_record" "em_fw_support_store_python_it_cname" {
   zone_id = aws_route53_zone.pythonit.id
-  name    = "em-fw.support.shop.python.it"
```
patrick91, line 421: I think this was wrong before 😊

PR #4077, opened 2024-09-29
Add printable version of the sponsorship brochure

patrick91:
Still very work in progress


backend/conferences/migrations/0048_conferencesponsorshippackageoption_and_more.py
```diff
@@ -0,0 +1,83 @@
+# Generated by Django 5.0.8 on 2024-09-29 15:07
```
patrick91, line 1: I need to squash the migrations

PR #4119, opened 2024-10-12
Migrate Grants vouchers to ConferenceVoucher

marcoacierno:
## What

Converts to use the Conference voucher system instead of a custom voucher system for grants.
The old system used to create the voucher on pretix and as well and the ability to send the email which we don't have automated in the Conference voucher system yet 

## ToDo

- [x] If a ConferenceVoucher already exists, and it is CO_SPEAKER we need to upgrade it to GRANT?

### DIFF
--- a/backend/conferences/models/conference_voucher.py
+++ b/backend/conferences/models/conference_voucher.py
@@ -1,6 +1,7 @@
 from django.db import models
 from django.utils.crypto import get_random_string
 from django.utils.translation import gettext_lazy as _
+from conferences.querysets import ConferenceVoucherQuerySet
 from model_utils.models import TimeStampedModel
 
 from conferences.models import Conference
@@ -48,6 +49,8 @@ class ConferenceVoucher(TimeStampedModel):
     voucher_email_sent_at = models.DateTimeField(
         help_text=_("When the email was last sent"), blank=True, null=True
     )
+
+    objects = ConferenceVoucherQuerySet().as_manager()
 
     def get_voucher_configuration(self):
         if self.voucher_type in (
--- a/backend/conferences/querysets.py
+++ b/backend/conferences/querysets.py
@@ -1,3 +1,6 @@
+from django.db import models
+
+
 class ConferenceQuerySetMixin:
     @property
     def conference_lookup_field(self):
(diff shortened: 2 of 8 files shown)


backend/conferences/querysets.py
```diff
@@ -8,3 +11,7 @@ def for_conference(self, conference):
 
     def for_conference_code(self, conference):
         return self.filter(**{f"{self.conference_lookup_field}__code": conference})
+
+
+class ConferenceVoucherQuerySet(ConferenceQuerySetMixin, models.QuerySet):
```
estyxx, line 16: Why crating a custom class ` ConferenceVoucherQuerySet` if does not add anything? Can't you just use `ConferenceQuerySetMixin` then?

backend/grants/admin.py
```diff
@@ -303,24 +291,47 @@ def create_grant_vouchers_on_pretix(modeladmin, request, queryset):
...
+                ConferenceVoucher(
+                    conference_id=grant.conference_id,
+                    user_id=grant.user_id,
+                    voucher_code=ConferenceVoucher.generate_code(),
+                    voucher_type=ConferenceVoucher.VoucherType.GRANT,
+                )
+            )
+            continue
+
+        if existing_voucher.voucher_type == ConferenceVoucher.VoucherType.CO_SPEAKER:
```
estyxx, line 313: What if the type is `SPEAKER`??

backend/conferences/models/conference_voucher.py
```diff
@@ -49,6 +50,8 @@ class VoucherType(models.TextChoices):
         help_text=_("When the email was last sent"), blank=True, null=True
     )
 
+    objects = ConferenceVoucherQuerySet().as_manager()
+
```
estyxx, line 54: Not the purpose of this PR, but why this configuration is hardcoded below?? 
if we change our mind, we need to update the code... what if conferences have different discounts? 🤔

backend/grants/admin.py
marcoacierno: If the type is speaker we already give them a 100% discount and I think it is more correct to keep their voucher marked as SPEAKER, as technically their ticket is already included in the speakers budget

backend/conferences/querysets.py
marcoacierno: ConferenceQuerySetMixin contains only the logic of the mixin, it isn't a `models.QuerySet`

backend/grants/admin.py
estyxx: oh I see, right!

estyxx (approved):
LGTM

PR #4168, opened 2024-11-15
Initial Brochure style fixes

marcoacierno:


patrick91:
@marcoacierno if you use divs instead of a table then the table header won't get duplicated when the table goes on multiple pages

PR #4177, opened 2024-11-24
Add confirmation email for grant application

estyxx:
## What

Sends a confirmation email to the applicant acknowledging receipt of their grant application.


backend/notifications/models.py
```diff
@@ -29,6 +29,10 @@ class EmailTemplateIdentifier(models.TextChoices):
 
     reset_password = "reset_password", _("[System] Reset password")
 
+    grant_application_confirmation = (
```
marcoacierno, line 32: Do you think this name is clear enough?

estyxx: any better ideas?

marcoacierno: can't think of a better name, "application" is what thrown me off a bit but it is not a big deal

marcoacierno (approved):
Once you merge this and it is deployed, can you create the template in admin?

PR #4290, opened 2025-01-01
Setup separate VPCs per each env

marcoacierno:


marcoacierno:
Takes around 6 mins to move the DB to a different VPC + we might need to replace the EC2 instance, so this needs to be merged after the CFP closes and over night

PR #4292, opened 2025-01-04
Serverless arm builder

marcoacierno:


marcoacierno:
Closing this for now because while it works, the reality is that once Github hosted ARM runners are finally on the free plan, I will switch to them: https://github.com/github/roadmap/issues/970 and given that they are planned to be releases 2025 Q1 (delayed from 2024 Q4 however) I feel like we can keep our current rough Hetzner setup for now

PR #4335, opened 2025-01-28
Remove voucher columns from Grant model

estyxx:
Removed the columns `pretix_voucher_id`, `voucher_code`, and
`voucher_email_sent_at` from the `Grant` model. These columns are no
longer needed as we now use the `ConferenceVoucher` table for voucher
management.


backend/grants/admin.py
```diff
@@ -461,9 +459,6 @@ class GrantAdmin(ExportMixin, ConferencePermissionMixin, admin.ModelAdmin):
                     "total_amount",
                     "applicant_reply_sent_at",
                     "applicant_reply_deadline",
-                    "pretix_voucher_id",
```
marcoacierno, line 464: It might be beneficial to expose in this admin something that checks if a ConferenceVoucher exists for this grant user, so we don't need to go to a different UI to check if this grant has the voucher or not, what do you think?

estyxx: yeah was thinking about.

estyxx: I added the columns `has_voucher` and `has_ticket`

backend/grants/admin.py
```diff
@@ -560,6 +561,31 @@ def emoji_gender(self, obj):
...
+            logger.error(e)
+            return None
+
+    @admin.display(
+        boolean=True,
+    )
+    def has_voucher(self, obj: Grant) -> bool:
+        if obj.has_voucher:
+            return True
+        return None
```
marcoacierno, line 587: Couldn't this just be `return obj.has_voucher`?

backend/grants/admin.py
```diff
@@ -560,6 +561,31 @@ def emoji_gender(self, obj):
...
+        if not obj.user_id:
+            return None
+
+        try:
+            return user_has_admission_ticket(
+                email=obj.user.email,
+                event_organizer=obj.conference.pretix_organizer_id,
+                event_slug=obj.conference.pretix_event_id,
+            )
+        except Exception as e:
```
marcoacierno, line 577: When does this happen?

estyxx: Locally I was getting this: 

```
403 Client Error: Forbidden for url: https://tickets.pycon.it/api/v1/organizers/python-italia/events/pycon12/tickets/attendee-has-ticket/
```

Probably i have something misconfigured... but can happen in theory also in prod?

marcoacierno: Hmm I don't think it can as prod? you should also probably update your local config to point to the right conference

estyxx: ok removed

PR #4339, opened 2025-01-31
Add Django debug toolbar for local development 

estyxx:


backend/pycon/urls.py
```diff
@@ -34,6 +34,14 @@
     path("", include("notifications.urls")),
 ] + static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
+
+if settings.DEBUG:
+    if settings.ENABLE_DJANGO_DEBUG_TOOLBAR:
```
marcoacierno, line 39: do we need the 2 nested `ifs`?

estyxx: changed in one line

marcoacierno (approved):

PR #4395, opened 2025-03-09
Store sent emails in S3

marcoacierno:


marcoacierno:
@claude can you update the PR so it abstracts the check if body_file is set or body is set without having to replicate it everywhere?

PR #4396, opened 2025-03-11
Send given and family name for invoice info

patrick91:
This PR changes from sending Full Name to Pretix to sending Given Name and Family Name


patrick91:
If I understood correctly codegen fails because it's trying to use the current backend 😊

marcoacierno (approved):

PR #4407, opened 2025-04-26
Add new SDI len label

libremente:
## What

Since #4406 has been merged, the old label does not hold anymore.
This is just a quick proposal, feel free to correct/ignore it. 
Thanks!


patrick91 (approved):
Thank you!

PR #4501, opened 2025-11-30
Add max length limitation to open questions and raise error for when travel fields are empty and travel is selected

estyxx:
- **Enforce travel detail fields for grant applications in backend**
- **Remove `django-debug-toolbar` from test config to fix local test failure**
- **Add character limitations for all open question fields**


marcoacierno:
@claude do a code review

PR #4582, opened 2026-02-14
Fix summary to handle pending status properly

estyxx:


estyxx:
@claude do the code review again

PR #4676, opened 2026-06-20
Remove blog posts

marcoacierno:
## What

We now use Wagtail for this kind of the marketing / website content

We are keeping the DB table for now in case we need some content from it in the future


marcoacierno:
@claude address the review

Medium: demodata/blog.json not deleted
This fixture references blog.post (a now-deleted model). Any workflow that loads all files in demodata/ will raise a DeserializationError. The file should be deleted alongside the app.

Low: blog still listed in pyproject.toml ruff isort config
/backend/pyproject.toml:31 has "blog" in the [tool.ruff.lint.isort] known-first-party list. The module no longer exists — dead config that should be removed.

